PR #2, opened 2022-09-16
[POAE7-2320] Support string as key/target in group by

Deegue:
Enable string support in Group by and add some tests.

Improvement.

No.

UTs.

cider, test

### DIFF
--- a/cider/exec/template/GroupByAndAggregate.cpp
+++ b/cider/exec/template/GroupByAndAggregate.cpp
@@ -323,21 +323,7 @@ GroupByAndAggregate::GroupByAndAggregate(
     , ra_exe_unit_(ra_exe_unit)
     , query_infos_(query_infos)
     , row_set_mem_owner_(row_set_mem_owner)
-    , group_cardinality_estimation_(group_cardinality_estimation) {
-  for (const auto& groupby_expr : ra_exe_unit_.groupby_exprs) {
-    if (!groupby_expr) {
-      continue;
-    }
-    const auto& groupby_ti = groupby_expr->get_type_info();
-    if (groupby_ti.is_bytes()) {
-      throw std::runtime_error(
-          "Cannot group by string columns which are not dictionary encoded.");
-    }
-    if (groupby_ti.is_buffer()) {
-      throw std::runtime_error("Group by buffer not supported");
-    }
-  }
-}
+    , group_cardinality_estimation_(group_cardinality_estimation) {}
 
 int64_t GroupByAndAggregate::getShardedTopBucket(const ColRangeInfo& col_range_info,
                                                  const size_t shard_count) const {
--- a/cider/tests/functionality/CiderGroupByTest.cpp
+++ b/cider/tests/functionality/CiderGroupByTest.cpp
@@ -50,24 +50,24 @@
 
 class CiderGroupByVarcharTest : public CiderTestBase {
  public:
+  // TODO(yizhong): make col_d to 50% chance nullable and change min length to 0
+  // after null string is supportedz
   CiderGroupByVarcharTest() {
     table_name_ = "table_test";
     create_ddl_ =
         "CREATE TABLE table_test(col_a BIGINT NOT NULL, col_b BIGINT NOT NULL, col_c "
         "VARCHAR NOT NULL, col_d VARCHAR);";
-    input_ = {std::make_shared<CiderBatch>(QueryDataGenerator::generateBatchByTypes(
-        500,
-        {"col_a", "col_b", "col_c", "col_d"},
-        {CREATE_SUBSTRAIT_TYPE(I64),
-         CREATE_SUBSTRAIT_TYPE(I64),
-         CREATE_SUBSTRAIT_TYPE(Varchar),
-         CREATE_SUBSTRAIT_TYPE(Varchar)},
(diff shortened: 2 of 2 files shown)


cider/exec/template/GroupByAndAggregate.cpp
```diff
@@ -323,21 +323,7 @@ GroupByAndAggregate::GroupByAndAggregate(
...
-    if (groupby_ti.is_bytes()) {
-      throw std::runtime_error(
-          "Cannot group by string columns which are not dictionary encoded.");
-    }
-    if (groupby_ti.is_buffer()) {
-      throw std::runtime_error("Group by buffer not supported");
-    }
-  }
-}
+    , group_cardinality_estimation_(group_cardinality_estimation) {}
```
winningsix, line 326: Move this to header this we do nothing in constructor.

cider/tests/functionality/CiderGroupByTest.cpp
```diff
@@ -48,6 +48,29 @@
     }                                                                                  \
   };
 
+class CiderGroupByVarcharTest : public CiderTestBase {
+ public:
+  // TODO(yizhong): make col_d to 50% chance nullable and change min length to 0
+  // after null string is supportedz
```
winningsix, line 54: typo: supportedz

winningsix (commented):
Hmm, there is nothing else we need to and just need to enable it by removing error throwing, right?

winningsix:
@bigPYJ1151 please help review this.

cider/exec/template/GroupByAndAggregate.cpp
jikunshang: No need to move. an empty constructor is a constructor impl, it's good to separate define and impl in .h and .cpp file.

cider/tests/functionality/CiderGroupByTest.cpp
```diff
@@ -440,6 +467,128 @@ TEST_F(CiderGroupByPrimitiveTypeMixTest, noConditionGroupByMultiColTest) {
       true);
 }
 
+TEST_F(CiderGroupByVarcharTest, varcharGroupByTest) {
+  /*one not null varchar group by key*/
+  assertQuery(
```
jikunshang, line 472: can you add some simple cases like this ? `SELECT SUM(col_a), col_c FROM table_test GROUP BY  col_c`

Deegue:
> Hmm, there is nothing else we need to and just need to enable it by removing error throwing, right?

Yes, we only need to enable it here since we have finished [[POAE7-2277]](https://github.com/intel-innersource/frameworks.ai.modular-sql.velox-plugin/pull/301) and [[POAE7-2278]](https://github.com/intel-innersource/frameworks.ai.modular-sql.velox-plugin/pull/318).

cider/tests/functionality/CiderGroupByTest.cpp
Deegue: added

cider/tests/functionality/CiderGroupByTest.cpp
Deegue: done

cider/tests/functionality/CiderGroupByTest.cpp
```diff
@@ -48,6 +48,29 @@
...
+        "CREATE TABLE table_test(col_a BIGINT NOT NULL, col_b BIGINT NOT NULL, col_c "
+        "VARCHAR NOT NULL, col_d VARCHAR);";
+    input_ = {std::make_shared<CiderBatch>(
+        QueryDataGenerator::generateBatchByTypes(500,
+                                                 {"col_a", "col_b", "col_c", "col_d"},
+                                                 {CREATE_SUBSTRAIT_TYPE(I64),
+                                                  CREATE_SUBSTRAIT_TYPE(I64),
+                                                  CREATE_SUBSTRAIT_TYPE(Varchar),
+                                                  CREATE_SUBSTRAIT_TYPE(Varchar)},
+                                                 {0, 0, 0, 0},
```
yma11, line 67: Don't you need set correct null_chance based on your defined schema?

Deegue: Commented a TODO at Line53, let me change the `null_chance` after null string is supported.

cider/tests/functionality/CiderGroupByTest.cpp
```diff
@@ -20,7 +20,7 @@
  */
 
 #include <gtest/gtest.h>
-#include "tests/utils/CiderTestBase.h"
+#include "CiderTestBase.h"
```
xieqi, line 23: as a practice, better to include the full file path, see [Names and Order of Includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

jikunshang: let's add `include_order` check in `cpplint` later.

jikunshang (approved):

Deegue:
Gentle ping @winningsix @yma11 for another review since this PR blocks many TPC-H queries like Q1, Q5, Q18 according to offline discussion with @harborn .

winningsix (approved):

PR #26, opened 2022-09-22
[POAE7-2372] Fix non-GroupBy avg with null value

Deegue:
Fix non-GroupBy avg when all values are null and add related UTs.

Bug fix.

No.

UTs.


Deegue:
Offline discussed, working in progress to cover fp cases.

cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -744,14 +744,32 @@ CiderBatch CiderRuntimeModule::setSchemaAndUpdateCountDistinctResIfNeed(
...
-        cast_buffer[0] = value;
+        if (target_infos[flatten_index].agg_arg_type.is_integer()) {
+          int64_t* result = const_cast<int64_t*>(
+              reinterpret_cast<const int64_t*>(outBuffers[flatten_index]));
+          double* cast_buffer = reinterpret_cast<double*>(result);
+          switch (result[0]) {
+            case std::numeric_limits<int8_t>::min():
+            case std::numeric_limits<int16_t>::min():
+            case std::numeric_limits<int32_t>::min():
+            case std::numeric_limits<int64_t>::min():
```
yma11, line 757: If `agg_arg_type.is_integer()`, the null value could only be null of `int64_t` and no other cases, since `sum(integer)->bigint`.

cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -744,14 +744,32 @@ CiderBatch CiderRuntimeModule::setSchemaAndUpdateCountDistinctResIfNeed(
...
+              break;
+            default:
+              cast_buffer[0] = (double)result[0];
+          }
+        } else if (target_infos[flatten_index].agg_arg_type.is_fp()) {
+          double* result = const_cast<double*>(
+              reinterpret_cast<const double*>(outBuffers[flatten_index]));
+          double* cast_buffer = reinterpret_cast<double*>(result);
+          if (result[0] == std::numeric_limits<float>::min()) {
+            cast_buffer[0] = NULL_VALUE_DOUBLE;
```
yma11, line 768: Similar as above, if `agg_arg_type.is_fp()`, the null value could only be `NULL_VALUE_DOUBLE` and no other cases and for this case, you should modify `[extract_column](https://github.com/intel/BDTK/blob/4f316da4e36b8646cbdf3add8d13b124ed4c94d1/cider/exec/module/CiderRuntimeModule.cpp#L390)`.

cider/exec/module/CiderRuntimeModule.cpp
Deegue: What if we got a substrait plan like in substrait-java keeps the same column type for sum output col as input col which doesn't meet our requirement? I think doing this will enhance our compatibility.

cider/exec/module/CiderRuntimeModule.cpp
Deegue: Get it, let me modify `extract_column` to transfer back null values of float.

cider/exec/module/CiderRuntimeModule.cpp
yma11: It's limited by Cider rule:
`    case SQLAgg::kSUM:
      return arg_expr->get_type_info().is_integer()
                 ? SQLTypeInfo(SQLTypes::kBIGINT, false)
                 : arg_expr->get_type_info();`
So here in the memory, there is no chance of tinyint/smallint/int/float null value exist.
For the case you are talking about, you should update `extract_column` for different null handle of different type, similar as float. And I am not sure whether this problem exists.

cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -744,14 +744,32 @@ CiderBatch CiderRuntimeModule::setSchemaAndUpdateCountDistinctResIfNeed(
       int flatten_index = schema->getFlattenColIndex(i);
       if (schema->getColHints()[i] == ColumnHint::PartialAVG &&
           target_infos[flatten_index].agg_kind == SQLAgg::kSUM &&
-          target_infos[flatten_index].agg_arg_type.is_integer() &&
           generator::getSQLTypeInfo(schema->getColumnTypeById(i).struct_().types(0))
```
jikunshang, line 747: 1. `schema->getColumnTypeById(i).struct_().types(0).has_fp64()` would be better?
2. what if this is not a double column?

cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -744,14 +744,32 @@ CiderBatch CiderRuntimeModule::setSchemaAndUpdateCountDistinctResIfNeed(
...
                   .get_type() == SQLTypes::kDOUBLE) {
-        int64_t* result = const_cast<int64_t*>(
-            reinterpret_cast<const int64_t*>(outBuffers[flatten_index]));
-        double value = (double)result[0];
-        double* cast_buffer = reinterpret_cast<double*>(result);
-        cast_buffer[0] = value;
+        if (target_infos[flatten_index].agg_arg_type.is_integer()) {
```
jikunshang, line 749: Please add necessary comments, it's hard to understand for others.

cider/exec/module/CiderRuntimeModule.cpp
jikunshang: and for L737, `may have` or `must have`?

jikunshang: And why do we do such process in this function `setSchemaAndUpdateCountDistinctResIfNeed`? I think it's confusing, we need do some refactor later.

yma11: > 1. `schema->getColumnTypeById(i).struct_().types(0).has_fp64()` would be better?
should be okay.
> 2. what if this is not a double column?
In this specific partial avg case, actual for velox, we need to convert bigint to double to match the output schema. In other case, like normal `sum`, the data should already be correct after extract, so we don't need special handle.

Deegue:
@jikunshang @yma11 Thanks for your review! All conversations above are resolved.

As for float and group by cases, it seems not easy to fix. Let me create another subtask and PR to handle them.

cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -734,24 +741,29 @@ CiderBatch CiderRuntimeModule::setSchemaAndUpdateCountDistinctResIfNeed(
 
   if (ciderCompilationResult_->impl_->query_mem_desc_->getQueryDescriptionType() ==
       QueryDescriptionType::NonGroupedAggregate) {
-    // for non-groupby partial avg, sum(int) may has double output Type thus
-    // need cast original int value(based on cider rule) to double
+    // For non-groupby partial avg, the output type of SUM(integer-type col) must be
```
jikunshang, line 744: my understanding here is the output type is decided by caller side generated plan. So Cider will not have a strong behavior here. in other words, only specific output types are supported.

cider/include/cider/CiderRuntimeModule.h
```diff
@@ -148,6 +147,8 @@ class CiderRuntimeModule {
   int32_t total_matched_rows_{0};
   int32_t fetched_rows_{0};
   constexpr static size_t kMaxOutputRows = 1000;
+  constexpr static double NULL_VALUE_DOUBLE = std::numeric_limits<double>::min();
+  constexpr static int32_t NULL_VALUE_BIGINT = std::numeric_limits<int32_t>::min();
```
jikunshang, line 151: int32_t for NULL_VALUE_BIGINT?

jikunshang: and rename to kBigIntNullValue?

cider/exec/module/CiderRuntimeModule.cpp
Deegue: yep, we change actual output data here to meet the requirement of Velox and there is only int64 which is needed to do this cast.

cider/include/cider/CiderRuntimeModule.h
Deegue: Renamed, sorry for that mistake.

yma11 (approved):
Overall LGTM now. One follow up is that please ping @ZJie1 for further velox integration.

jikunshang (approved):

PR #35, opened 2022-09-27
[POAE7-2368][Hot Fix] Workaround for Nullability Problem Under NOT IS NULL Expression.

bigPYJ1151:

Workaround for Nullability Problem Under NOT IS NULL Expression.

Plan from Velox will always notes nullability as nullable. However, IS NULL expression should be not nullable, which will lead to codegen problem under old data format.

No.

UTs.

Bug

Pls take a review @Ma-Jian1

### DIFF
--- a/cider/exec/template/LogicalIR.cpp
+++ b/cider/exec/template/LogicalIR.cpp
@@ -483,7 +483,13 @@ llvm::Value* CodeGenerator::codegenLogical(const Analyzer::UOper* uoper,
   CHECK(operand_ti.is_boolean());
   const auto operand_lv = codegen(operand, true, co).front();
   CHECK(operand_lv->getType()->isIntegerTy());
-  const bool not_null = (operand_ti.get_notnull() || is_qualified_bin_oper(operand));
+  bool not_null = (operand_ti.get_notnull() || is_qualified_bin_oper(operand));
+  // Workaround for isNull expr.
+  if (auto operand_uoper = dynamic_cast<const Analyzer::UOper*>(operand)) {
+    if (kISNULL == operand_uoper->get_optype()) {
+      not_null = true;
+    }
+  }
   CHECK(not_null || operand_lv->getType()->isIntegerTy(8));
   return not_null
              ? cgen_state_->ir_builder_.CreateNot(toBool(operand_lv))


Ma-Jian1 (approved):

cider/exec/template/LogicalIR.cpp
```diff
@@ -483,7 +483,13 @@ llvm::Value* CodeGenerator::codegenLogical(const Analyzer::UOper* uoper,
   CHECK(operand_ti.is_boolean());
   const auto operand_lv = codegen(operand, true, co).front();
   CHECK(operand_lv->getType()->isIntegerTy());
-  const bool not_null = (operand_ti.get_notnull() || is_qualified_bin_oper(operand));
+  bool not_null = (operand_ti.get_notnull() || is_qualified_bin_oper(operand));
+  // Workaround for isNull expr.
```
winningsix, line 487: Do we need a TODO here?

cider/exec/template/LogicalIR.cpp
```diff
@@ -483,7 +483,13 @@ llvm::Value* CodeGenerator::codegenLogical(const Analyzer::UOper* uoper,
...
+  if (auto operand_uoper = dynamic_cast<const Analyzer::UOper*>(operand)) {
+    if (kISNULL == operand_uoper->get_optype()) {
+      not_null = true;
+    }
```
yma11, line 491: Seems originally `not_null` is determined by `UOper`'s argument, why do we need to fix it to true?

cider/exec/template/LogicalIR.cpp
bigPYJ1151: TODO what? 🤔

cider/exec/template/LogicalIR.cpp
bigPYJ1151: The problem was encountered in E2E Test. Due to nullability of expression from Velox always is nullable.

cider/exec/template/LogicalIR.cpp
winningsix: It's mentioning workaround.

winningsix (approved):

PR #38, opened 2022-09-28
[POAE7-2366] between...and... string support(cast str to date tpch-q7 workaround)

kd97:

between...and... str support

cast str to date workaround to fix tpch q7.

No

UTs

BUG

### DIFF
--- a/cider/tests/functionality/CiderStringTest.cpp
+++ b/cider/tests/functionality/CiderStringTest.cpp
@@ -77,7 +77,7 @@ class CiderStringToDateTest : public CiderTestBase {
     table_name_ = "test";
     create_ddl_ = R"(CREATE TABLE test(col_int INTEGER, col_str VARCHAR(10));)";
     input_ = {std::make_shared<CiderBatch>(QueryDataGenerator::generateBatchByTypes(
-        50,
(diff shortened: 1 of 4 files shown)


cider/tests/functionality/CiderStringTest.cpp
```diff
@@ -86,6 +86,10 @@ class CiderStringToDateTest : public CiderTestBase {
 };
 
 TEST_F(CiderStringToDateTest, DateStrTest) {
+  assertQuery(
+      "select col_str from test where col_str between date '1970-01-01' and date "
+      "'2077-12-31'",
+      "cast_str_to_date_implictly.json");
```
yma11, line 92: How about adding a test with null covered?

kd97: class `DateStrTest `will generate  varchar type null value with a 50% chance

https://github.com/intel/BDTK/blob/d46dfd8703ab19b023b1c52d3096289cd24bad8a/cider/tests/functionality/CiderStringTest.cpp#L76-L85

yma11 (approved):

winningsix (approved):

PR #40, opened 2022-09-29
[POAE7-2378] Add support for using non-encoded string's function in filter

spevenhe:

Add string function translation support for non-encoded string 
The updates are based on the latest heavydb code.

to support using non-encoded string's function in filter(bin_op) and other ops.

No

Uts

BUG


cider/exec/template/ScalarExprVisitor.h
```diff
@@ -147,6 +143,10 @@ class ScalarExprVisitor {
     if (agg) {
       return visitAggExpr(agg);
     }
+    const auto string_oper = dynamic_cast<const Analyzer::StringOper*>(expr);
```
jikunshang, line 146: will this be called? can you point out code path?

spevenhe: mainly in substraitToEu's filter part

ctx_element->accept(rel_visitor_ptr);
=>
FilterQualContext::accept
=>
FilterRelVisitor::visit
=>
and after substrait2Analyzer (toAnalyzerExpr)
=>
fold_expr(cider_expr.get())
=>
visitor.visit()

jikunshang (approved):

winningsix (approved):

PR #41, opened 2022-09-29
[POAE7-2370] Incorrect string conversion

xieqi:

Fix incorrect string conversion due to slot offset

incorrect offset for extract column

No

Using exist UTs

cider

### DIFF
--- a/cider/exec/module/CiderRuntimeModule.cpp
+++ b/cider/exec/module/CiderRuntimeModule.cpp
@@ -63,6 +63,9 @@ using agg_query_hoist_literals = void (*)(const int8_t***,  // col_buffers
                                           const int64_t*);  // join_hash_tables_ptr
 namespace {
 
+constexpr static double kDoubleNullValue = std::numeric_limits<double>::min();
+constexpr static int64_t kBigIntNullValue = std::numeric_limits<int64_t>::min();
+
 template <typename T>
 std::enable_if_t<std::is_integral<T>::value, void> extract_column(int32_t start_row,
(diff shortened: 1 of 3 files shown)


cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -317,172 +418,87 @@ void CiderRuntimeModule::processNextBatch(const CiderBatch& in_batch) {
...
-                 EIGHT_BYTES;
-  }
-
   auto schema = ciderCompilationResult_->getOutputCiderTableSchema();
   if (!outBatch.schema()) {
     outBatch.set_schema(std::make_shared<CiderTableSchema>(schema));
   }
+
+  const auto& query_mem_desc = ciderCompilationResult_->impl_->query_mem_desc_;
+  const auto col_slot_context = query_mem_desc->getColSlotContext();
```
jikunshang, line 438: can you point out what kind of case will original offset not work well?

jikunshang (commented):
LGTM.

cider/exec/module/CiderRuntimeModule.cpp
xieqi: the getColOffInBytes is not correct when column has more than one slot, e.g. non-encoding sting, so the offset will always be 1,2,3,4,5....

harborn (approved):

harborn (approved):

harborn (approved):

winningsix (approved):

PR #48, opened 2022-09-30
[POAE7-2380] [HotFix] Groupby HashTable Code Refactor and Bugfix

bigPYJ1151:

- Remove keyless and columnar-layout related code in CiderAggHashTable.
- Fix offset bug under group-by w/o aggregations.
- Fix mistake introduced by previous conflict resolve.

Bug fix.

No

UTs

BugFix

Pls take a review @harborn @winningsix

### DIFF
--- a/cider/exec/operator/aggregate/CiderAggHashTable.cpp
+++ b/cider/exec/operator/aggregate/CiderAggHashTable.cpp
@@ -252,9 +252,6 @@ CiderAggHashTable::CiderAggHashTable(
                        : (query_mem_desc->getGroupbyColCount() +
                           query_mem_desc->getBufferColSlotCount()))
     , key_columns_num_(query_mem_desc->getKeyCount())
-    , is_keyless_(query_mem_desc->hasKeylessHash())
-    , keyless_key_index_(query_mem_desc->getTargetIdxForKey())
-    , is_columnar_layout_(query_mem_desc->didOutputColumnar())
     , row_width_(query_mem_desc->getRowSize())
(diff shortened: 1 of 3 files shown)


cider/exec/operator/aggregate/CiderAggHashTable.cpp
```diff
@@ -312,7 +307,7 @@ std::vector<CiderAggHashTableEntryInfo> CiderAggHashTable::fillColsInfo() {
   std::vector<CiderAggHashTableEntryInfo> cols_info(columns_num_);
 
   size_t offset = 0, col_slot_context_index = 0;
-  bool is_key = !is_keyless_;
+  bool is_key = true;
```
Deegue, line 310: Do we always have one or more keys here no matter it is in count or avg?

Deegue (commented):
LGTM. Just one question above, and I'm not sure why we should remove is_keyless. Maybe it's also related to Arrow format?

harborn (approved):

bigPYJ1151:
> LGTM. Just some questions, and I'm not sure why we should remove is_keyless. Maybe it's also related to Arrow format?

keyless hash is a trivial optimization which is not in use, so just remove it.

cider/exec/operator/aggregate/CiderAggHashTable.cpp
bigPYJ1151: For group-by there will be one key col at least.

winningsix (approved):

PR #54, opened 2022-10-09
[POAE7-2388] change join hash table building to single thread and update previous tests

spevenhe:

1.  change join hash table building to single thread from cpu_threads
2.  updates hash join tests

1. As a library, cider should not use all the threads.
2.  Previous join tests had problems so updates

No

Uts

### DIFF
--- a/cider/exec/template/operator/join/hashtable/BaselineJoinHashTable.cpp
+++ b/cider/exec/template/operator/join/hashtable/BaselineJoinHashTable.cpp
@@ -284,7 +284,7 @@ std::pair<size_t, size_t> BaselineJoinHashTable::approximateTupleCount(
               << ", emitted keys count: " << cached_count_info.second;
       return std::make_pair(*cached_count_info.first, cached_count_info.second);
     }
-    int thread_count = cpu_threads();
+    int thread_count = 1;
     std::vector<uint8_t> hll_buffer_all_cpus(thread_count * padded_size_bytes);
     auto hll_result = &hll_buffer_all_cpus[0];
 
--- a/cider/tests/functionality/CiderHashJoinTest.cpp
+++ b/cider/tests/functionality/CiderHashJoinTest.cpp
@@ -272,7 +272,6 @@ HASH_JOIN_TEST_UNIT(CiderOneToOneSeqJoinTest, ExprJoinSeqTest2, *, d, = 2 *)
 HASH_JOIN_TEST_UNIT(CiderOneToOneSeqJoinTest, ExprJoinSeqTest3, *, d, *2 = 2 *)
 
 // using OR to avoid 0 results
-// TODO: (spevenhe) OR will fail back to loop join while AND is still hash join
 #define DOUBLE_JOIN_OR_CONDITION_TEST(TEST_CLASS, UNIT_NAME, PROJECT)               \
   TEST_F(TEST_CLASS, UNIT_NAME) {                                                   \
     /*INNER JOIN ON INTEGER OR FLOAT*/                                              \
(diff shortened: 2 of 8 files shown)


cider/tests/functionality/CiderHashJoinTest.cpp
```diff
@@ -272,24 +272,64 @@ HASH_JOIN_TEST_UNIT(CiderOneToOneSeqJoinTest, ExprJoinSeqTest2, *, d, = 2 *)
...
+    assertJoinQueryRowEqualAndReset(                                                \
+        "SELECT " #PROJECT                                                          \
+        " from table_probe JOIN table_hash ON l_b = r_b OR l_b IS NOT NULL ");      \
+    /*LEFT JOIN ON BIGINT AND NOT NULL*/                                            \
+    assertJoinQueryRowEqualAndReset(                                                \
+        "SELECT " #PROJECT                                                          \
+        " from table_probe LEFT JOIN table_hash ON l_a = r_a OR l_a IS NOT NULL "); \
+  }
+
+// DOUBLE_JOIN_OR_CONDITION_TEST(CiderOneToOneSeqJoinTest, ORJoinConditionTest1, *)
```
jikunshang, line 304: why disable these?

jikunshang (approved):

cider/tests/utils/CiderTestBase.cpp
```diff
@@ -105,6 +105,7 @@ void CiderJoinTestBase::assertJoinQuery(const std::string& sql,
 void CiderJoinTestBase::assertJoinQueryRowEqual(const std::string& sql,
                                                 const std::string& json_file,
                                                 const bool ignore_order) {
+  VLOG(4) << sql;
```
yma11, line 108: Need revert?

cider/tests/functionality/CiderHashJoinTest.cpp
spevenhe: Unlike ```AND``` which uses hash join, conditions with ```OR``` will fall back to loop join. Haven't investigated the reason yet

cider/tests/utils/CiderTestBase.cpp
spevenhe: During generating EU, I see lots of log in omnisci's code, so I added one to debug. Do we prefer no log in test code?

cider/tests/functionality/CiderHashJoinTest.cpp
winningsix: Please add some comments why we comment this out.

cider/exec/template/operator/join/hashtable/BaselineJoinHashTable.cpp
```diff
@@ -284,7 +284,7 @@ std::pair<size_t, size_t> BaselineJoinHashTable::approximateTupleCount(
               << ", emitted keys count: " << cached_count_info.second;
       return std::make_pair(*cached_count_info.first, cached_count_info.second);
     }
-    int thread_count = cpu_threads();
+    int thread_count = 1;
```
winningsix, line 287: Hmm, can we make this a configurable option (e.g., a config in https://github.com/intel/BDTK/blob/eda3ea46087b9ba1893f5d8e520814ca4281fca4/cider/include/cider/CiderOptions.h#L105)instead of hardcode?

jikunshang: I think it's not necessary.

spevenhe: As a database, multi-thread makes senses. But for cider, I thought it just should be one thread.

jikunshang (approved):

cider/exec/template/operator/join/hashtable/BaselineJoinHashTable.cpp
winningsix: It's common to have multi-threading hash table builder. But current usage of Cider is quite limited (requiring all data ready).  It's a boarder discussion. Let's have a follow-up discussion with @chaojun-zhang.

Back to this PR. LGTM to merge this first.

winningsix (approved):

PR #57, opened 2022-10-11
[POAE7-2181] Fix string error in batch checker and add tests for IS DISTINCT/IS NOT DISTINCT

Deegue:
Remove empty string value initialization in constructor of batch checker.
Add string and mixed test cases and improve comments.

Fix & Improvement.

No.

UTs.


winningsix (dismissed):
Seems adding test for " IS DISTINCT/IS NOT DISTINCT" and fix string error are irrelevant? If so, we should create two PRs next time.

cider/util/sqldefs.h
```diff
@@ -26,7 +26,8 @@
 // in sync.
 enum SQLOps {
   kEQ = 0,
-  // BitWise Equal <=> a = b or (a is null and b is null)
+  // The only difference between EQ and BW_EQ is that BW_EQ will check nullability of both
```
jikunshang, line 29: `check nullability of both sides` means `(a is null and b is null)` ?

jikunshang (approved):

cider/util/sqldefs.h
yma11: Seems previous comments is more clear. They should both check nullability but for `kEQ`, it doesn't return true if both sides are null?

Deegue:
> Seems adding test for " IS DISTINCT/IS NOT DISTINCT" and fix string error are irrelevant? If so, we should create two PRs next time.

Yes, this test exposed the bug. I will separate them next time.

cider/util/sqldefs.h
Deegue: See [check-equal-bitwise](https://iq.opengenus.org/check-equal-bitwise/) for more information.

jikunshang (approved):

yma11 (approved):

PR #60, opened 2022-10-12
[POAE7-2333] Arrow based batch test framework

jikunshang:

1. reuse current cider batch data generator, and convert to arrow based batch in test. will not convert null data currently.
2. for result check, just compare row/column num with duck result, didn't verify actual data.
3. will replace with new utils.

### DIFF
--- a/cider/exec/module/batch/CiderBatch.cpp
+++ b/cider/exec/module/batch/CiderBatch.cpp
@@ -253,6 +253,51 @@ void CiderBatch::releaseArrowEntries() {
   }
 }
 
+// to be deprecated, just for test not nullable data.
+void CiderBatch::convertToArrowRepresentation() {
+  CHECK(!arrow_array_ && !arrow_schema_);
+  arrow_array_ = CiderBatchUtils::allocateArrowArray();
+  arrow_schema_ = CiderBatchUtils::allocateArrowSchema();
+
+  arrow_array_->length = row_num();
+  arrow_array_->n_children = column_num();
+  arrow_array_->buffers = nullptr;  // ?
+  arrow_array_->n_buffers = 0;      // ?
+  arrow_array_->private_data = nullptr;
+  arrow_array_->children = (ArrowArray**)std::malloc(sizeof(ArrowArray) * column_num());
+  arrow_array_->release = CiderBatchUtils::ciderEmptyArrowArrayReleaser;
+
+  arrow_schema_->format = "+s";
+  arrow_schema_->dictionary = nullptr;
+  arrow_schema_->n_children = column_num();
+  arrow_schema_->children =
+      (ArrowSchema**)std::malloc(sizeof(ArrowSchema*) * column_num());
+  arrow_schema_->release = CiderBatchUtils::ciderEmptyArrowSchemaReleaser;
+
+  for (int i = 0; i < column_num(); i++) {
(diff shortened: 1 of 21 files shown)


jikunshang:
@bigPYJ1151 @xieqi @yma11 Please take a review.

cider/exec/module/batch/CiderBatch.cpp
```diff
@@ -253,6 +253,51 @@ void CiderBatch::releaseArrowEntries() {
   }
 }
 
+// to be deprecated, just for test not nullable data.
+void CiderBatch::convertToArrowRepresentation() {
```
spevenhe, line 257: This is converting to structbatch right? if column_num() is 1, should ciderbatch convert to strcutbatch or scalarbatch?

jikunshang: yes. I think even a single column batch should convert to structBatch.

spevenhe: Then what is the use of scalarbatch?

jikunshang: I think codegen internal will never use scalarBatch. maybe it's just for test. Correct me if I am wrong. @bigPYJ1151

cider/tests/utils/CiderTestBase.cpp
```diff
@@ -41,6 +41,23 @@ void CiderTestBase::assertQuery(const std::string& sql,
...
 
+void CiderTestBase::prepareArrowBatch() {
+  for (auto iter = input_.begin(); iter != input_.end(); iter++) {
+    (*iter)->convertToArrowRepresentation();
+  }
+}
+
+void CiderTestBase::assertQueryArrow(const std::string& sql,
+                                     const std::string& json_file) {
+  auto duck_res = duckDbQueryRunner_.runSql(sql);
```
yma11, line 52: currently duckdb input is still converted based on original format, so we also need to update the convertor based on arrow format right?

jikunshang: yes, borui is working on this.

cider/exec/template/NativeCodegen.cpp
```diff
@@ -1896,8 +1896,9 @@ bool Executor::compileBody(const RelAlgExecutionUnit& ra_exe_unit,
       filter_lv = cgen_state_->ir_builder_.CreateAnd(
           filter_lv, cgen_state_->ir_builder_.CreateNot(is_null));
     }
-    return cgen_state_->ir_builder_.CreateAnd(filter_lv,
-                                              fixedsize_expr_value->getValue());
+    // need to convert to bool expr first.
+    auto cond = code_generator.toBool(fixedsize_expr_value->getValue());
```
bigPYJ1151, line 1900: Confused, results of filter related expressions (logicalOp, cmpOP, .etc) should be bool type in my mind. Any special cases here?

jikunshang: constant need this. query like `SELECT col_1 FROM test WHERE 2 = 2`.

yma11 (approved):
I am okay to merge this so that we can start using the CiderBatchToArrow util for more migration.

bigPYJ1151 (approved):

PR #67, opened 2022-10-13
[POAE7-2400] DuckDB output to Arrow-format CiderBatch

YBRua:

Updated `DuckDbResultConvertor` to support `ArrowArray` and `ArrowSchema`. Currently supports *primitive scalar types*.

1. In `DuckDbResultConvertor`: Added `fetchDataToArrowFormattedCiderBatch` method. The API use case is the same as the existing `fetchDataToCiderBatch`, but it returns `CiderBatch` instances that uses Arrow format as underlying memory layout.
2. In `DuckDbQueryRunnerTest.cpp`: Added simple handcrafted tests on new convertor API for primitive types (`int` / `float` / `double` / `bool`), along with util functions to run these tests. Newly added tests are put under `DuckDBResultConvertorTest`.
   - Unit tests here are subject to change. They will be updated when `CiderBatchChecker` are updated to support Arrow format.

These changes are part of the changes needed to migrate unit test framework to Arrow Format.

Example use case of the API:

```cpp
/* check cider/tests/utils/DuckDbQueryRunnerTest.cpp for details */

/* create table and insert data into runner */
DuckDbQueryRunner runner;
// ...

auto res = runner.runSql("select * from table_test;");
// actual_batches will contain `CiderBatch`es with Arrow format support.
auto actual_batches =
        DuckDbResultConvertor::fetchDataToArrowFormattedCiderBatch(res);

```

### DIFF
--- a/cider/tests/utils/DuckDbQueryRunner.cpp
+++ b/cider/tests/utils/DuckDbQueryRunner.cpp
@@ -470,6 +470,77 @@ void addColumnDataToScalarBatch<CiderByteArray>(
   child->setNullCount(null_count);
 }
 
+CiderBatch DuckDbResultConvertor::fetchOneArrowFormattedBatch(
+    std::unique_ptr<duckdb::DataChunk>& chunk) {
+  // Construct ArrowSchema
+  // First build a CiderTableSchema and then convert it
+  std::vector<::substrait::Type> batch_types;
+  std::vector<std::string> batch_names;
+
+  int col_num = chunk->ColumnCount();
+  int row_num = chunk->size();
+  std::vector<::duckdb::LogicalType> types = chunk->GetTypes();
+  for (int i = 0; i < col_num; ++i) {
+    auto type = types[i];
+    batch_types.emplace_back(std::move(convertToSubstraitType(type)));
+    batch_names.emplace_back("");
+    if (!isDuckDb2CiderBatchSupportType(type)) {
+      CIDER_THROW(CiderCompileException,
+                  "Non numberic type " + type.ToString() + " not supported yet");
+    }
+  }
+
+  auto cider_schema = std::make_shared<CiderTableSchema>(batch_names, batch_types, "");
+  auto arrow_schema =
+      CiderBatchUtils::convertCiderTableSchemaToArrowSchema(*cider_schema);
+
+  // Construct ArrowArray
+  /// NOTE: DuckDb provides an API DataChunk::ToArrowArray()
+  /// for exporting data to ArrowArray, but we are NOT using it
+  /// since the structs for DATE and VARCHAR differ from those used in DuckDB
+  auto batch =
+      StructBatch::Create(arrow_schema, std::make_shared<CiderDefaultAllocator>());
+  for (int i = 0; i < col_num; ++i) {
+    auto child = batch->getChildAt(i);
+    auto type = types[i];
+    switch (type.id()) {
+      case ::duckdb::LogicalTypeId::HUGEINT:
+      case ::duckdb::LogicalTypeId::BIGINT:
+      case ::duckdb::LogicalTypeId::TIMESTAMP:
+      case ::duckdb::LogicalTypeId::TIME:
+        addColumnDataToScalarBatch<int64_t>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::INTEGER:
+        addColumnDataToScalarBatch<int32_t>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::SMALLINT:
+        addColumnDataToScalarBatch<int16_t>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::BOOLEAN:
+      case ::duckdb::LogicalTypeId::TINYINT:
+        addColumnDataToScalarBatch<int8_t>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::FLOAT:
+        addColumnDataToScalarBatch<float>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::DOUBLE:
+      case ::duckdb::LogicalTypeId::DECIMAL:
+        addColumnDataToScalarBatch<double>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::DATE:
+        addColumnDataToScalarBatch<CiderDateType>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::CHAR:
+      case ::duckdb::LogicalTypeId::VARCHAR:
+        addColumnDataToScalarBatch<CiderByteArray>(chunk, i, row_num, child.get());
+        break;
+      default:
+        CIDER_THROW(CiderCompileException,
+                    "Unsupported type convertor: " + type.ToString());
+    }
+  }
+}
+
 CiderBatch DuckDbResultConvertor::fetchOneBatch(
     std::unique_ptr<duckdb::DataChunk>& chunk) {
   const int col_num = chunk->ColumnCount();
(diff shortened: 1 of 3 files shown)


cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -388,6 +390,164 @@ void addColumnDataToCiderBatch<CiderByteArray>(
...
+                                int i,
+                                int row_num,
+                                CiderBatch* child) {
+  CHECK(child->resizeBatch(row_num, true));
+  auto data_buffer = child->asMutable<ScalarBatch<T>>()->getMutableRawData();
+  auto null_buffer = child->getMutableNulls();
+  int64_t null_count = 0;
+
+  for (int j = 0; j < row_num; j++) {
+    T value = std::numeric_limits<T>::min();  // init with Cider null value
```
jikunshang, line 404: we don't need to set a cider null value currently, 0 would be fine.

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -388,6 +390,164 @@ void addColumnDataToCiderBatch<CiderByteArray>(
...
+      value = dataChunk->GetValue(i, j).GetValue<T>();
+      CiderBitUtils::setBitAt(null_buffer, j);
+    }
+    std::memcpy(data_buffer + j, &value, sizeof(T));
+  }
+  child->setNullCount(null_count);
+}
+
+template <>
+void addColumnDataToScalarBatch<CiderDateType>(
```
jikunshang, line 418: I think you can remove this since `CiderDateType` will be replaced by Arrow date representation. You can add this method in later PRs. Same as L443 one.

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -388,6 +390,164 @@ void addColumnDataToCiderBatch<CiderByteArray>(
...
+                                int row_num,
+                                CiderBatch* child) {
+  CHECK(child->resizeBatch(row_num, true));
+  auto data_buffer = child->asMutable<ScalarBatch<T>>()->getMutableRawData();
+  auto null_buffer = child->getMutableNulls();
+  int64_t null_count = 0;
+
+  for (int j = 0; j < row_num; j++) {
+    T value = std::numeric_limits<T>::min();  // init with Cider null value
+    if (dataChunk->GetValue(i, j).IsNull()) {
```
jikunshang, line 405: I noticed that `DataChunk` has such method `ToArrowArray`, maybe we can use this.

cider/tests/utils/DuckDbQueryRunner.cpp
YBRua: OK. So both `CiderDateType` and `CiderByteArray` should be removed right?

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -388,6 +390,164 @@ void addColumnDataToCiderBatch<CiderByteArray>(
...
+      buf[j].ptr = value_buf;
+      CiderBitUtils::setBitAt(null_buffer, j);
+    }
+  }
+  child->setNullCount(null_count);
+}
+
+CiderBatch DuckDbResultConvertor::fetchOneArrowFormattedBatch(
+    std::unique_ptr<duckdb::DataChunk>& chunk) {
+  // Construct ArrowSchema
```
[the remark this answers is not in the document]
YBRua: I didn't realize `CiderTableSchema` would also be removed in the future. I'll fix this.

I chose to convert a `CiderTableSchema` because existing functions only allowed DuckDb types to be converted to Substrait types, but we could not directly build an `ArrowSchema` from Substrait types, and I wanted to avoid adding another type conversion function. But since CiderTableSchema will also be deprecated, seems a conversion is inevitable.

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -388,6 +390,164 @@ void addColumnDataToCiderBatch<CiderByteArray>(
...
+        addColumnDataToScalarBatch<int64_t>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::INTEGER:
+        addColumnDataToScalarBatch<int32_t>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::SMALLINT:
+        addColumnDataToScalarBatch<int16_t>(chunk, i, row_num, child.get());
+        break;
+      case ::duckdb::LogicalTypeId::BOOLEAN:
+        addColumnDataToScalarBatch<bool>(chunk, i, row_num, child.get());
```
spevenhe, line 523: may be put these in a order from small to large, eg bool, tiny, small, int....

YBRua: These lines have been removed. We are using methods provided by DuckDb to directly construct an ArrowArray.

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -388,6 +395,47 @@ void addColumnDataToCiderBatch<CiderByteArray>(
...
+  auto arrow_array = CiderBatchUtils::allocateArrowArray();
+  chunk->ToArrowArray(arrow_array);
+
+  // Build CiderBatch
+  auto batch = StructBatch::Create(
+      arrow_schema, std::make_shared<CiderDefaultAllocator>(), arrow_array);
+
+  // ToArrowArray() will not compute null_count, so we do it manually here
+  for (int i = 0; i < col_num; ++i) {
+    auto child = batch->getChildAt(i);
```
jikunshang, line 420: make L420-433 as a method?

cider/tests/utils/DuckDbQueryRunner.h
```diff
@@ -60,15 +62,28 @@ class DuckDbQueryRunner {
...
+  /// Will be appended as-is after the colon character.
+  static std::vector<std::shared_ptr<CiderBatch>> fetchDataToArrowFormattedCiderBatch(
+      std::unique_ptr<::duckdb::MaterializedQueryResult>& result,
+      std::string config_timezone = "");
 
  private:
   static CiderBatch fetchOneBatch(std::unique_ptr<duckdb::DataChunk>& chunk);
+  static CiderBatch fetchOneArrowFormattedBatch(std::unique_ptr<duckdb::DataChunk>& chunk,
+                                                std::vector<std::string>& names,
+                                                std::string& config_timezone);
```
jikunshang, line 86: what about `std::string config_timezone = ""` ?

cider/tests/utils/DuckDbQueryRunner.cpp
YBRua: It is now encapsulated into a private member function `updateChildrenNullCounts`. But I'm not sure if this name is proper.

cider/tests/utils/DuckDbQueryRunner.h
YBRua: updated

jikunshang (approved):
LGTM :)

yma11:
@YBRua can you retrigger new round of CI/CD? seems build failed but not sure about the root cause.

winningsix:
> @YBRua can you retrigger new round of CI/CD? seems build failed but not sure about the root cause.

It may be due to query runners are killed during lab maintenance.

cider/tests/utils/DuckDbQueryRunnerTest.cpp
```diff
@@ -136,7 +412,7 @@ TEST(DuckDBQueryRunnerTest, insertCiderBatchTest) {
   auto res = runner.runSql("select * from table_test;");
   CHECK(!res->HasError());
   CHECK_EQ(res->ColumnCount(), 2);
-  std::cout << "error: " << res->error << std::endl;
+  std::cout << "error: " << res->error << std::endl;  // YBRua: wrap this into an if?
```
yma11, line 415: @jikunshang should we replace all std::cout using the `CiderException`?

yma11: +1.

YBRua: this line has been removed

yma11 (approved):

jikunshang (approved):

PR #68, opened 2022-10-13
[POAE7-2408] Replace folly::split of function look up

haiweizh:

folly is not used by debug model in function look up, that will cause compilation failure, fix that use cider util split replace folly:split

folly is not used by debug model in function look up, that will cause compilation failure, fix that use cider util split replace folly:split

No

Uts

BUG


haiweizh:
Bug fix. Plz help to take a review. @yma11 @jikunshang @harborn

harborn (approved):

haiweizh:
https://jira.devtools.intel.com/browse/POAE7-2408

xieqi (approved):

PR #37, opened 2022-09-27
[POAE7-2376]add logger doc

PenghuiJiao:
please take review @winningsix @xieqi @Ma-Jian1 
### What changes were proposed in this pull request?

add logger doc

### Why are the changes needed?

add logger doc

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

No
### Which label does this PR belong to?

cider

PR #70, opened 2022-10-14
[POAE7-2432] Fix Old CiderBatch Memory Allocation for Group-By Result Fetching

bigPYJ1151:

Previous memory allocation of output old CiderBatch doesn't follow the schema in CiderOutputTableSchema or CiderAggHashTable, which works fine when the size of item less than 8 bytes. However, the size of VarChar type recently supported is 16 bytes, it leads memory problems.

No

UTs and e2e test.

BugFix

Pls take a review @harborn @winningsix


bigPYJ1151:
#37 didn't pass the UTs but was merged, I have fixed it. Pls take review carefully and make the CI really in use.

harborn (approved):

winningsix (approved):

PR #72, opened 2022-10-17
[POAE7-2435] fix col name

Ma-Jian1:
fix bug in col name

bug

No

UTs

BUG

### DIFF
--- a/cider/exec/plan/parser/PlanRelVisitor.cpp
+++ b/cider/exec/plan/parser/PlanRelVisitor.cpp
@@ -33,11 +33,11 @@ AggRelVisitor::AggRelVisitor(const substrait::AggregateRel& rel_node,
                              Substrait2AnalyzerExprConverter* toAnalyzerExprConverter,
                              const std::unordered_map<int, std::string>& function_map,
                              std::shared_ptr<VariableContext> variable_context_shared_ptr,
-                             bool is_right_join_node)
+                             bool is_join_right_node)
     : RelVisitor(toAnalyzerExprConverter,
                  function_map,
                  variable_context_shared_ptr,
-                 is_right_join_node)
+                 is_join_right_node)
     , rel_node_(rel_node) {}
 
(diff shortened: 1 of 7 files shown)


cider/exec/plan/parser/PlanRelVisitor.cpp
```diff
@@ -431,15 +433,15 @@ void ReadRelVisitor::visit(InputDescContext* input_desc_context) {
   for (int i = 0; i < rel_node_.base_schema().struct_().types_size(); i++) {
     auto type = rel_node_.base_schema().struct_().types(i);
     auto sti = getSQLTypeInfo(type);
-    ColumnInfoPtr colum_info_ptr =
-        std::make_shared<ColumnInfo>(100, cur_table_id, i, "col_" + i, sti, false);
+    ColumnInfoPtr colum_info_ptr = std::make_shared<ColumnInfo>(
+        fake_db_id, cur_table_id, i, fmt::format("col_{}", i), sti, false);
```
yma11, line 437: When will `"col_" + i` not work and how does it looks like?

Ma-Jian1: `"col_" + i` means `"col_"[i]`

yma11: Verified. Yes we should wrap this `i` with `std::to_string.

cider/exec/plan/parser/PlanRelVisitor.cpp
```diff
@@ -24,6 +24,8 @@
  * @brief   Plan Rel visitor
  **/
 
+#include <fmt/core.h>
```
yma11, line 27: remove this?

Ma-Jian1: done

bigPYJ1151 (dismissed):

bigPYJ1151 (approved):

PR #75, opened 2022-10-18
[POAE7-2410] CiderBatchChecker on primitive types

YBRua:

Added `checkArrowEq()` API in `CiderBatchChecker`. This method is designed to be used in the Arrow-based test framework. It checks if two vectors of Arrow-formatted `CiderBatch` instances are equal. The checking logic and API signature is the same as the old `checkEq()` API.

Several notes here (also commented in codes)

1. The checker now supports all primitive scalar types **excluding boolean values** (to be more exact, it currently supports integers and floating points). Support for boolean is postponed, because the memory layout of boolean ArrowArrays (bytes or bit-packed values) is yet to be determined.
2. The checker now **DOES NOT** check `ArrowSchema`, because the arrow schema in `CiderBatch` is a private member, and its getter function is currently a *protected* method. The checker has no access to the schema. However, if a `CiderBatch` is constructed correctly, the type and number of columns of the underlying arrow array should be consistent with the schema, so it should suffice to only check the array, I suppose.
3. Old temporary implementation of `checkArrowEq` is now renamed as `checkArrowEqTemp` to avoid signature conflicts.

Migrating Cider unit test framework to Arrow format.

No

Tested with hand-crafted unit tests. More robust unit tests are to be updated, after Builder and Generator are updated.


cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -21,6 +21,367 @@
...
+
+  int row_num = actual_batch->getLength();
+
+  // compare nulls
+  bool null_buffer_eq = checkValidityBitmapEqual(expected_batch, actual_batch);
+  if (!null_buffer_eq) {
+    /// NOTE: (YBRua) even if check fails, the result is not necessarily incorrect
+    /// because padding values in the same byte can be different and fails memcmp
+    /// e.g. 01011111 and 00011111 are both correct when row_num == 5
+    std::cout << "Null buffer compare failed." << std::endl;
```
yma11, line 244: Why don't we compare this null_vec one by one bit to avoid the scenario run into here?

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -21,6 +21,367 @@
...
+    return false;
+  }
+
+  // compare data
+  bool data_buffer_eq = true;
+  data_buffer_eq = !memcmp(expected_data_buffer, actual_data_buffer, row_num * sizeof(T));
```
yma11, line 250: In arrow format, `null` is represented by null_vec and the value of `data_buffer[i]` in corresponding position is meaningless but here we compare these positions in `memcmp`, maybe we should remove this step of compare and leverage `ConcatenatedRow` directly? @jikunshang

cider/tests/utils/CiderBatchChecker.cpp
YBRua: I wasn't doing it because this entire function was used for comparing memory chunks using `memcmp`, and doing a bitwise operation in this context looked somewhat weird. But it is indeed a bad idea to check null buffer with `memcmp` here because it is very likely to fail.

YBRua: I think I'll update this after we decide whether this `memcmp`-based step is still necessary. I'll either change it into a bitwise op, or directly remove it.

cider/tests/utils/CiderBatchChecker.cpp
jikunshang: if both null buffer and data buffer are equal, we can bypass later check. if not, will do later check using `ConcatenatedRow`.

YBRua: I think the point @yma11 is trying to make is that the `memcmp`-based check is _more likely to fail_ than it was in the old `checkEq` implementation (not sure if I got it right). IMO, the checking logic here will work in some cases, but it is going to fail more often than before, so I am not sure whether to keep it.

yma11: It's okay if the memcmp check fail since then we will have `ConcatenatedRow` check eventually even under this scenario it's redundant. One open is that whether its 100% reliable if returns a `true` result after checking the values including meaningless ones.

YBRua: It (theoretically) returns true only if `null_buffer` and `data_buffer` both exactly match. In this case I don't think it will cause problems, even if the result contains null values.

Deegue: This part looks good, but I think here we no longer need one-to-one batch check(`checkOneStructBatchEqual` and `checkOneScalarBatchEqual`) due to the reason commented below.

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -21,6 +21,367 @@
...
+
+std::vector<ConcatenatedRow> CiderBatchChecker::arrowToConcatenatedRowVector(
+    const std::vector<std::shared_ptr<CiderBatch>>& cider_batches) {
+  std::vector<ConcatenatedRow> total_row;
+  for (auto batch : cider_batches) {
+    auto col_num = batch->getChildrenNum();
+
+    for (int row_index = 0; row_index < batch->getLength(); row_index++) {
+      auto batch_str =
+          CiderBatchStringifier::stringifyStructBatchAt(batch.get(), row_index);
```
jikunshang, line 165: I guess performance here will be bad, maybe we can construct typed Stringfier before loop

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -21,6 +21,367 @@
...
+}
+
+bool CiderBatchChecker::checkValidityBitmapEqual(const CiderBatch* expected_batch,
+                                                 const CiderBatch* actual_batch) {
+  auto expected_buffer = expected_batch->getNulls();
+  auto actual_buffer = actual_batch->getNulls();
+  auto expected_count = expected_batch->getNullCount();
+  auto actual_count = actual_batch->getNullCount();
+
+  if (expected_count != actual_count) {
```
jikunshang, line 209: in some case null count will not be set. this should not be a strong check.

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -21,6 +21,367 @@
...
+    std::cout << "expected_null_count != actual_null_count. "
+              << "Expected: " << expected_count << ". Actual: " << actual_count
+              << std::endl;
+    return false;
+  }
+
+  int buffer_size = (expected_count + 7) >> 3;
```
jikunshang, line 216: should be `row_count` ?

jikunshang (commented):
Overall LGTM.

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -21,6 +21,367 @@
...
+bool CiderBatchChecker::colNumCheck(
+    const std::vector<std::shared_ptr<CiderBatch>>& batches,
+    int expected_col_num) {
+  bool passed = true;
+  for (auto i = 0; i < batches.size(); ++i) {
+    if (batches[i]->getChildrenNum() != expected_col_num) {
+      std::cout << "Batch " << i << " is not having the same number of cols. "
+                << "Expected " << expected_col_num << "; Got "
+                << batches[i]->getChildrenNum() << std::endl;
+      passed = false;
```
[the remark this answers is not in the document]
YBRua: updated

cider/tests/utils/CiderBatchChecker.cpp
YBRua: A new `CheckBitVectorEq` method is implemented, and null_buffer is now checked with this method.

cider/util/CiderBitUtils.h
```diff
@@ -183,6 +183,27 @@ inline size_t countSetBits(const uint8_t* bit_vector, size_t end) {
 
   return ans;
 }
+
+inline bool CheckBitVectorEq(const uint8_t* vec1, const uint8_t* vec2, int end) {
+  size_t i = 0, bytes = 0;
+
+  // do byte-batched check with memcmp
+  for (; i + 64 <= end; i += 64, bytes++)
```
jikunshang, line 191: why 64? should be 8? and `bytes = end/8` would be fine?
1 byte  = 8 bits

YBRua: My fault. Copied directly from `countSetBits()` without thinking. `countSetBits()` did a `uint64_t` cast and used 64. But should be 8 here. Fixed.

YBRua:
Wait, still buggy. Fixing.

jikunshang (approved):

spevenhe (approved):
LGTM

cider/tests/utils/CiderBatchChecker.h
```diff
@@ -118,24 +126,50 @@ class CiderBatchChecker {
...
+  // 3. memcmp check: use memcmp if both vectors contain only one batch
+  //    which is more efficient and works as a shortcut
+  // 4. Full check: if previous check fails, or cannot be applied,
+  //    convert inputs to vectors of ConcatenatedRow and check with hashing
+  static bool checkArrowEq(
+      const std::vector<std::shared_ptr<CiderBatch>>& expected_batches,
+      const std::vector<std::shared_ptr<CiderBatch>>& actual_batches,
+      const bool ignore_order = false);
+
+  static bool checkArrowEq(std::shared_ptr<CiderBatch> expected_batch,
```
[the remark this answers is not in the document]
YBRua: Yes. These tests will be added after CiderBatchBuilder and QueryDataGenerator are updated, because it is a bit troublesome to manually create one-to-many and many-to-one tests cases.

cider/util/CiderBitUtils.h
```diff
@@ -183,6 +183,25 @@ inline size_t countSetBits(const uint8_t* bit_vector, size_t end) {
...
+
+inline bool CheckBitVectorEq(const uint8_t* vec1, const uint8_t* vec2, int end) {
+  // do byte-batched check with memcmp
+  size_t bytes = end / 8;
+  if (memcmp(vec1, vec2, bytes)) {
+    return false;
+  }
+
+  // compare remainders
+  size_t i = bytes * 8;
```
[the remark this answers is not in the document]
YBRua: Yes, thanks for reminding that there is a manual alignment and initialization in [ArrowArrayBuilder](https://github.com/intel/BDTK/blob/main/cider/tests/utils/ArrowArrayBuilder.h#L104). I didn't notice this earlier.

However, since this function is currently used for comparing null buffers, I expected the parameter `end` to be the number of rows in an arrow array, which may not be aligned to 8.

Further, I don't think this alignment and initialization can always be guaranteed, because if one creates the null buffer with `CiderBatch`'s `resizeBatch()` (which will call [resizeNulls()](https://github.com/intel/BDTK/blob/main/cider/tests/utils/ArrowArrayBuilder.h#L104) to resize null buffer), while the allocated memory is aligned to 8, there may be several remaining bits that are not initialized and can be any value (because `resizeNulls()` only initializes the first `size` bits, which may not necessarily align to 8). In this case if we directly `memcmp` the entire byte it is very likely to fail due to the uninitialized bits.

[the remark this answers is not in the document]
spevenhe: make sense. One more thing, arrow array uses [least-significant bit (LSB) numbering]. This means that within a group of 8 bits, it should read right-to-left:

https://arrow.apache.org/docs/format/Columnar.html#encapsulated-message-format

jikunshang:
Thanks for contribution!

PR #77, opened 2022-10-19
[POAE7-2454] ::duckdb::QueryResult::GetConfigTimezone not found when CIDER_ENABLE_DUCKDB enabled

YBRua:

Bug fix. In `DuckDbResultConvertor` we used duckdb's `ToArrowSchema()` method for exporting Arrow schemas. However, the versions of duckdb when `CIDER_ENABLE_VELOX` is enabled or disabled are different, and contain breaking API changes, the signature of `ToArrowSchema()` are slightly different, which leads to compile errors when `CIDER_ENABLE_VELOX=OFF`.

Since the two versions of duckdb differs in several APIs and it is troublesome to directly fix the issue, we extracted the logic of `ToArrowSchema()` from the duckdb source code and maintain a copy of this function independently (see `DuckDbArrowAdaptor.h`).

NOTE: setting `CIDER_ENABLE_VELOX=OFF` will not compile due to another error in `SubstraitExtension.cpp` (msg: error: cannot convert 'const YAML::Node' to 'bool' in return), a temporary fix to make the code at least compilable would be changing the return statement to `return false;` (`true` should also work).

Bug fix, as is described above.

No.

Re-compiled and ran UTs.

BUG


cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -416,15 +411,14 @@ void DuckDbResultConvertor::updateChildrenNullCounts(CiderBatch& batch) {
...
-    std::string config_timezone) {
+    std::vector<std::string>& names) {
   int col_num = chunk->ColumnCount();
   int row_num = chunk->size();
 
   // Construct ArrowSchema using methods from duckdb
   std::vector<::duckdb::LogicalType> types = chunk->GetTypes();
   auto arrow_schema = CiderBatchUtils::allocateArrowSchema();
-  ::duckdb::QueryResult::ToArrowSchema(arrow_schema, types, names, config_timezone);
+  DuckDbArrowSchemaAdaptor::duckdbResultToArrowSchema(arrow_schema, types, names);
```
kd97, line 422: can move `updateChildrenNullCounts` into this method and update inside?

YBRua: I'm afraid the answer is no. `null_count` is an attribute of arrow **array**, but we are dealing with arrow **schema** here, and arrow schema does not have a `null_count` attribute.

cider/tests/utils/DuckDbArrowAdaptor.h
```diff
@@ -0,0 +1,77 @@
...
+  static void SetArrowMapFormat(DuckDBArrowSchemaHolder& root_holder,
+                                ArrowSchema& child,
+                                const ::duckdb::LogicalType& type);
+
+  static void SetArrowFormat(DuckDBArrowSchemaHolder& root_holder,
+                             ArrowSchema& child,
+                             const ::duckdb::LogicalType& type);
+
+ public:
+  static void duckdbResultToArrowSchema(ArrowSchema* out_schema,
```
jikunshang, line 72: better rename to `duckdbResultSchemaToArrowSchema`

jikunshang (approved):

yma11 (approved):

jikunshang (dismissed):

jikunshang:
please fix code style issue.

jikunshang (dismissed):

yma11 (approved):

yma11 (approved):

jikunshang (approved):

PR #85, opened 2022-10-20
[POAE7-2416] In Ops using Arrow Format

yma11:
Switch InValues codegen to Arrow format

Arrow migration

No

UT

### DIFF
--- a/cider/exec/template/CodeGenerator.h
+++ b/cider/exec/template/CodeGenerator.h
@@ -224,7 +224,7 @@ class CodeGenerator {
   // Cider Data Format
   std::unique_ptr<CodegenColValues> codegenInValues(const Analyzer::InValues* expr,
                                                     const CompilationOptions& co);
-  // will deprecate
+  // TODO: (yma11) will deprecate
   llvm::Value* codegen(const Analyzer::InValues*, const CompilationOptions&);
 
   llvm::Value* codegen(const Analyzer::InIntegerSet* expr, const CompilationOptions& co);
--- a/cider/exec/template/CodeGenerator.h
+++ b/cider/exec/template/CodeGenerator.h
@@ -169,6 +169,14 @@ class CodeGenerator {
                           const Analyzer::Expr*,
                           const CompilationOptions&);
 
+  std::unique_ptr<CodegenColValues> codegenCmpFun(const SQLOps,
+                                                  const SQLQualifier,
+                                                  llvm::Value*,
+                                                  llvm::Value*,
+                                                  const SQLTypeInfo&,
+                                                  const Analyzer::Expr*,
+                                                  const CompilationOptions&);
+
   // Deprecating
   llvm::Value* codegenIsNull(const Analyzer::UOper*, const CompilationOptions&);
 
@@ -213,6 +221,10 @@ class CodeGenerator {
                            const bool operand_is_const,
                            const CompilationOptions& co);
 
+  // Cider Data Format
+  std::unique_ptr<CodegenColValues> codegenInValues(const Analyzer::InValues* expr,
+                                                    const CompilationOptions& co);
+  // will deprecate
   llvm::Value* codegen(const Analyzer::InValues*, const CompilationOptions&);
 
   llvm::Value* codegen(const Analyzer::InIntegerSet* expr, const CompilationOptions& co);
--- a/cider/exec/template/CompareIR.cpp
+++ b/cider/exec/template/CompareIR.cpp
@@ -477,6 +477,46 @@ llvm::Value* CodeGenerator::codegenCmpDecimalConst(const SQLOps optype,
   return codegenCmp(optype, qualifier, {lhs_lv}, new_ti, new_rhs_lit.get(), co);
 }
 
+// For Cider Data Format
+std::unique_ptr<CodegenColValues> CodeGenerator::codegenCmpFun(
+    const SQLOps optype,
+    const SQLQualifier qualifier,
+    llvm::Value* lhs_lv,
+    llvm::Value* lhs_lv_null,
+    const SQLTypeInfo& lhs_ti,
+    const Analyzer::Expr* rhs,
+    const CompilationOptions& co) {
+  AUTOMATIC_IR_METADATA(cgen_state_);
+  CHECK(IS_COMPARISON(optype));
+  const auto& rhs_ti = rhs->get_type_info();
+  auto rhs_lv = codegen(rhs, co, true);
+  auto rhs_lv_fixedsize = dynamic_cast<FixedSizeColValues*>(rhs_lv.get());
+  CHECK_EQ(kONE, qualifier);
+  CHECK((lhs_ti.get_type() == rhs_ti.get_type()) ||
+        (lhs_ti.is_string() && rhs_ti.is_string()));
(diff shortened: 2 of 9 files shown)


cider/function/scalar/RuntimeFunctions.cpp
```diff
@@ -394,6 +394,23 @@ extern "C" ALWAYS_INLINE int8_t bit_is_set(const int64_t bitset,
              : 0;
 }
 
+extern "C" ALWAYS_INLINE int8_t bit_is_set_cider(const int64_t bitset,
```
jikunshang, line 397: possible to pass a pointer instead?

yma11: Do you mean the `bitset`? it's the address of the bitmap and the bit check is based on interpreting it, as `reinterpret_cast<const int8_t*>(bitset))[bitmap_idx >> 3` so I think no need to pass a pointer.

jikunshang: as a function, it would be better to use int8_t* as input parameter which is more friendly to its user. There may be some limitation in code gen stage (like you cannnot pass a pointer directly when call this external function), so you do like this, right?

yma11: Understand your point. But for this function's use scenario, it's more straight to pass this `int64_t` address directly because the pattern is fixed, as well as the `int64_t val, val_min, val_max`.

jikunshang: Please refer `check_bit_vector_set` `check_bit_vector_clear` in RuntimeFunction.cpp

yma11: The bit check logic inside function `check_bit_vector_set ` is based on interpreting the `bit_vector` as `uint8_t` and that's make it consistent. In my user scenario, the bit set logic do 'reinterpret_cast<const int8_t*>' inside the function but for the function parameters, it uses int64_t. I think there is no big difference.

yma11: @jikunshang Thanks for your advice. Per offline discussion, we keep function like this for the time being and will consider reimplement it when switch to more efficient Bitmap late.

cider/tests/functionality/CiderFilterTest.cpp
```diff
@@ -509,6 +508,10 @@ TEST_F(CiderFilterRandomTestBase, DistinctFromTest) {
 
 int main(int argc, char** argv) {
   testing::InitGoogleTest(&argc, argv);
+  logger::LogOptions log_options(argv[0]);
```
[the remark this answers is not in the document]
yma11: This is to easily enable debugging options when do tests. It's more straight to add these lines here. We will update all test files eventually.

cider/function/scalar/RuntimeFunctions.cpp
jikunshang, line 397: will this function be called in current test cases? and return bool would be better?

cider/exec/template/InValuesIR.cpp
```diff
@@ -75,6 +76,53 @@ llvm::Value* CodeGenerator::codegen(const Analyzer::InValues* expr,
...
+    const CompilationOptions& co) {
+  AUTOMATIC_IR_METADATA(cgen_state_);
+  const auto in_arg = expr->get_arg();
+  if (is_unnest(in_arg)) {
+    CIDER_THROW(CiderCompileException, "IN not supported for unnested expressions");
+  }
+  const auto& expr_ti = expr->get_type_info();
+  CHECK(expr_ti.is_boolean());
+  auto lhs_lv = codegen(in_arg, co, true);
+  auto lhs_fixsize = dynamic_cast<FixedSizeColValues*>(lhs_lv.get());
```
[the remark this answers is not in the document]
yma11: added.

cider/exec/template/CompareIR.cpp
```diff
@@ -477,6 +477,46 @@ llvm::Value* CodeGenerator::codegenCmpDecimalConst(const SQLOps optype,
...
+  const auto& rhs_ti = rhs->get_type_info();
+  auto rhs_lv = codegen(rhs, co, true);
+  auto rhs_lv_fixedsize = dynamic_cast<FixedSizeColValues*>(rhs_lv.get());
+  CHECK_EQ(kONE, qualifier);
+  CHECK((lhs_ti.get_type() == rhs_ti.get_type()) ||
+        (lhs_ti.is_string() && rhs_ti.is_string()));
+  if (lhs_ti.is_integer() || lhs_ti.is_decimal() || lhs_ti.is_time() ||
+      lhs_ti.is_timeinterval()) {
+    // TODO: (yma1) need handle string/boolean
+    if (lhs_ti.is_string() || (lhs_ti.is_boolean() && rhs_ti.is_boolean())) {
```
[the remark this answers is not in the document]
yma11: Yes. Let me move it out of this check.

cider/exec/template/InValuesIR.cpp
```diff
@@ -75,6 +76,53 @@ llvm::Value* CodeGenerator::codegen(const Analyzer::InValues* expr,
...
+  auto lhs_value = lhs_fixsize->getValue();
+  auto lhs_null = lhs_fixsize->getNull();
+  llvm::Value* result{nullptr};
+  if (expr_ti.get_notnull()) {
```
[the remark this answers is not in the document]
yma11: It's original existing here, and not sure whether exists scenarios need this check.

cider/function/scalar/RuntimeFunctions.cpp
yma11: done.

cider/exec/template/CodeGenerator.h
```diff
@@ -213,6 +221,10 @@ class CodeGenerator {
                            const bool operand_is_const,
                            const CompilationOptions& co);
 
+  // Cider Data Format
+  std::unique_ptr<CodegenColValues> codegenInValues(const Analyzer::InValues* expr,
+                                                    const CompilationOptions& co);
+  // will deprecate
```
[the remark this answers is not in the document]
yma11: added.

yma11:
@winningsix @jikunshang PR rebased with comments addressed. Please take a review again. Thanks.

winningsix (approved):

cider/exec/template/CompareIR.cpp
```diff
@@ -477,6 +477,46 @@ llvm::Value* CodeGenerator::codegenCmpDecimalConst(const SQLOps optype,
...
+            llvm_icmp_pred(optype), lhs_lv, rhs_lv_fixedsize->getValue()),
+        lhs_lv_null);
+  }
+  if (lhs_ti.get_type() == kFLOAT || lhs_ti.get_type() == kDOUBLE) {
+    return std::make_unique<FixedSizeColValues>(
+        cgen_state_->ir_builder_.CreateFCmp(
+            llvm_fcmp_pred(optype), lhs_lv, rhs_lv_fixedsize->getValue()),
+        lhs_lv_null);
+  }
+  CHECK(false);
```
jikunshang, line 515: throw exception instead?

jikunshang (approved):

PR #90, opened 2022-10-24
[POAE7-2459] DuckDb and CiderChecker UT migration to Arrow format

YBRua:

Migrated all supported UTs in DuckDbQueryRunner and CiderBatchChecker to Arrow format.

- All currently supported test cases in DuckDbQueryRunnerTest.cpp and CiderBatchCheckerTest.cpp have been migrated.
- Test cases are now built with `ArrowArrayBuilder` and checked with `CiderBatchChecker::checkArrowEq`.
- Supported inserting Arrow-formatted data into DuckDb databases
- Fixed and improved several corner-case handling in `CiderBatchChecker` according to UTs.
  - Added nullptr check in `checkOneScalarBatchEqual`.
  - Raised exceptions if the buffer in ArrowArray is a nullptr.
  - Fixed check logic when only one of the null buffers is a nullptr.

Part of the UT framework migration to Apache Arrow format.

No.

UTs.

### DIFF
--- a/cider/tests/CiderBitUtilsTest.cpp
+++ b/cider/tests/CiderBitUtilsTest.cpp
@@ -102,18 +102,23 @@ TEST_F(CiderBitUtilsTest, CountTest) {
   auto bit_vec = bit_vec_1.as<uint8_t>();
 
   EXPECT_EQ(countSetBits(bit_vec, 1025), 0ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1025ul);
 
   setBitAt(bit_vec, 0);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 1ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1024ul);
 
   setBitAt(bit_vec, 1024);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 2ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1023ul);
 
   setBitAt(bit_vec, 1025);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 2ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1023ul);
 
   setBitAt(bit_vec, 100);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 3ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1022ul);
 }
 
 TEST_F(CiderBitUtilsTest, CheckEqTest) {
--- a/cider/tests/CiderBitUtilsTest.cpp
+++ b/cider/tests/CiderBitUtilsTest.cpp
@@ -102,23 +102,18 @@ TEST_F(CiderBitUtilsTest, CountTest) {
   auto bit_vec = bit_vec_1.as<uint8_t>();
 
   EXPECT_EQ(countSetBits(bit_vec, 1025), 0ul);
-  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1025ul);
 
   setBitAt(bit_vec, 0);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 1ul);
-  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1024ul);
 
   setBitAt(bit_vec, 1024);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 2ul);
-  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1023ul);
 
   setBitAt(bit_vec, 1025);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 2ul);
-  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1023ul);
 
   setBitAt(bit_vec, 100);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 3ul);
-  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1022ul);
 }
 
 TEST_F(CiderBitUtilsTest, CheckEqTest) {
--- a/cider/tests/utils/CiderBatchChecker.cpp
+++ b/cider/tests/utils/CiderBatchChecker.cpp
@@ -113,6 +113,15 @@ bool CiderBatchChecker::checkValidityBitmapEqual(const CiderBatch* expected_batc
 template <typename T>
 bool CiderBatchChecker::checkOneScalarBatchEqual(const ScalarBatch<T>* expected_batch,
                                                  const ScalarBatch<T>* actual_batch) {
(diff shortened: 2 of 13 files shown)


cider/tests/CiderBitUtilsTest.cpp
```diff
@@ -102,18 +102,23 @@ TEST_F(CiderBitUtilsTest, CountTest) {
...
   EXPECT_EQ(countSetBits(bit_vec, 1025), 0ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1025ul);
 
   setBitAt(bit_vec, 0);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 1ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1024ul);
 
   setBitAt(bit_vec, 1024);
   EXPECT_EQ(countSetBits(bit_vec, 1025), 2ul);
+  EXPECT_EQ(countUnsetBits(bit_vec, 1025), 1023ul);
```
yma11, line 113: Based on definition of `countUnsetBits`, seems it's not necessary to this check, I mean, 1023 is always equal to 1025 plus 2.

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -134,6 +134,45 @@ ::duckdb::Value duckTimestampValueAt(const int8_t* buf, int64_t offset) {
   return timestamp;
 }
 
+template <typename T>
+::duckdb::Value duckValueAtScalarBatch(const ScalarBatch<T>* batch, int64_t offset) {
```
yma11, line 138: rename to `duckdbValueAtScalarBatch`?

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -243,6 +282,41 @@ void DuckDbQueryRunner::createTableAndInsertData(
...
+    int col_num = current_batch->getChildrenNum();
+    for (int row_idx = 0; row_idx < row_num; ++row_idx) {
+      ::duckdb::Appender appender(con, table_name);
+      appender.BeginRow();
+      for (int col_idx = 0; col_idx < col_num; ++col_idx) {
+        auto child = current_batch->getChildAt(col_idx);
+        auto child_type = child->getCiderType();
+        auto valid_bitmap = child->getNulls();
+        if (!valid_bitmap || CiderBitUtils::isBitSetAt(valid_bitmap, row_idx)) {
+          auto value = GEN_DUCK_VALUE_FROM_ARROW_FUNC();
```
yma11, line 309: rename to `GEN_DUCK_DB_VALUE_FROM_ARROW_FUNC`.

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -243,6 +282,41 @@ void DuckDbQueryRunner::createTableAndInsertData(
...
+    int row_num = current_batch->getLength();
+    int col_num = current_batch->getChildrenNum();
+    for (int row_idx = 0; row_idx < row_num; ++row_idx) {
+      ::duckdb::Appender appender(con, table_name);
+      appender.BeginRow();
+      for (int col_idx = 0; col_idx < col_num; ++col_idx) {
+        auto child = current_batch->getChildAt(col_idx);
+        auto child_type = child->getCiderType();
+        auto valid_bitmap = child->getNulls();
+        if (!valid_bitmap || CiderBitUtils::isBitSetAt(valid_bitmap, row_idx)) {
```
yma11, line 308: I suggest for duckdb converter, we don't leverage the `null_vector` as the check step but directly do value cp to keep consistent with input values for cider.

cider/tests/CiderBitUtilsTest.cpp
YBRua: Yes. Should I add some different test cases or should I just remove these?

cider/tests/utils/DuckDbQueryRunner.cpp
YBRua: But if we directly copy all values and skip nulls, DuckDb will not be able to tell which values are null and which values are valid.

cider/tests/CiderBitUtilsTest.cpp
yma11: I am not sure what you want to cover by these TCs. At least as current code, I think you can remove it.

cider/tests/utils/DuckDbQueryRunner.cpp
yma11: Okay. Got u.

yma11 (dismissed):

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -70,35 +70,58 @@ int CiderBatchChecker::getTotalNumOfRows(
...
     // both buffers are nullptr, no need to check
     return true;
   } else if (expected_buffer && actual_buffer) {
     // both buffers exist
     return CiderBitUtils::CheckBitVectorEq(expected_buffer, actual_buffer, row_num);
+  } else {
+    // one is nullptr but the other is not, null_count of the other batch must be 0
+    if (!actual_null_cnt) {
+      // actual values are all valid, expected null count should be 0
+      expected_null_cnt = expected_null_cnt == -1
```
jikunshang, line 99: 1. default null_count is `-1`?
2. should be `CiderBitUtils::countSetBits(null_buffer)` ?

jikunshang (commented):
Overall LGTM.

cider/tests/utils/CiderBatchChecker.cpp
YBRua: 1. `null_count` may be `-1` if it is not yet computed ([Apache Arrow doc](https://arrow.apache.org/docs/format/CDataInterface.html#c.ArrowArray.null_count)). But it seems in the current arrow array builder computing `null_count` has not yet been implemented (marked as a TODO), and in current implementation the initial uncomputed value for `null_count` is `0`.
2. Actually an `ArrowArray` does not have a "null buffer". It only has a "validity bitmap", where `1` represents non-null and `0` represents null. It is the inverse of a null buffer, so we should count unset bits (0's, i.e. null values) in the validity bitmap. (The name `getNulls()` is actually a bit misleading because it actually gets a validity bitmap instead of a null bitmap).

YBRua:
The previous commit was reverted, because initializing `null_count` as `-1` broke Unit Tests in CiderArrowBatchTest, and it was difficult to fix the issue without breaking the consistency and logic of existing code (both in CiderBatch implementation and in the unit test).

`null_count` remains zero-initialized. It will sometimes cause a false alert in `checkOneScalarBatchEqual`, but it will not affect the overall correctness as long as the validity bitmap is correctly set (because we can still check the result with `ConcatenatedRow`)

winningsix (approved):

yma11 (approved):

PR #97, opened 2022-10-24
[POAE7-2417] Support NotNull Ops for Arrow Format and some fix

Deegue:
Some fixes and improvements to make it an easier way to use new Arrow Format including data generator and checker.

1. Fix empty result arrow batch core dump.
2. Fix compile error when using both data generator.
3. Fix and use `checkArrowEq` in `assertQueryArrow`.
4. Fix arrow format in `SetUp` of `CiderTestBase`.
5. Add and enable tests for null and not null.

fix & improvements.

No.

UTs.

### DIFF
--- a/cider/tests/functionality/CiderFilterTest.cpp
+++ b/cider/tests/functionality/CiderFilterTest.cpp
@@ -25,6 +25,343 @@
 
 #include "tests/utils/CiderTestBase.h"
 
+// To be deperacated. old test cases.
+class CiderFilterSequenceTestBase : public CiderTestBase {
+ public:
+  CiderFilterSequenceTestBase() {
+    table_name_ = "test";
+    create_ddl_ =
+        R"(CREATE TABLE test(col_1 INTEGER, col_2 BIGINT,
+        col_3 FLOAT, col_4 DOUBLE);)";
+    input_ = {std::make_shared<CiderBatch>(
+        QueryDataGenerator::generateBatchByTypes(99,
+                                                 {"col_1", "col_2", "col_3", "col_4"},
+                                                 {CREATE_SUBSTRAIT_TYPE(I32),
+                                                  CREATE_SUBSTRAIT_TYPE(I64),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp32),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp64)}))};
+  }
+};
+
+class CiderFilterRandomTestBase : public CiderTestBase {
+ public:
+  CiderFilterRandomTestBase() {
+    table_name_ = "test";
+    create_ddl_ =
+        R"(CREATE TABLE test(col_1 INTEGER, col_2 BIGINT, col_3 FLOAT, col_4 DOUBLE,
+           col_5 INTEGER, col_6 BIGINT, col_7 FLOAT, col_8 DOUBLE, col_9 VARCHAR(10), col_10 VARCHAR(10));)";
+    input_ = {std::make_shared<CiderBatch>(
+        QueryDataGenerator::generateBatchByTypes(999,
+                                                 {"col_1",
+                                                  "col_2",
+                                                  "col_3",
+                                                  "col_4",
+                                                  "col_5",
+                                                  "col_6",
+                                                  "col_7",
+                                                  "col_8",
+                                                  "col_9",
+                                                  "col_10"},
+                                                 {CREATE_SUBSTRAIT_TYPE(I32),
+                                                  CREATE_SUBSTRAIT_TYPE(I64),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp32),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp64),
+                                                  CREATE_SUBSTRAIT_TYPE(I32),
+                                                  CREATE_SUBSTRAIT_TYPE(I64),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp32),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp64),
+                                                  CREATE_SUBSTRAIT_TYPE(Varchar),
+                                                  CREATE_SUBSTRAIT_TYPE(Varchar)},
+                                                 {2, 2, 2, 2, 2, 2, 2, 2, 2, 2},
+                                                 GeneratePattern::Random,
+                                                 1,
+                                                 100))};
+  }
+};
+
+class CiderFilterNullTestBase : public CiderTestBase {
+ public:
+  CiderFilterNullTestBase() {
+    table_name_ = "test";
+    create_ddl_ =
+        R"(CREATE TABLE test(col_1 INTEGER, col_2 BIGINT,
+        col_3 FLOAT, col_4 DOUBLE);)";
+    input_ = {std::make_shared<CiderBatch>(
+        QueryDataGenerator::generateBatchByTypes(99,
+                                                 {"col_1", "col_2", "col_3", "col_4"},
+                                                 {CREATE_SUBSTRAIT_TYPE(I32),
+                                                  CREATE_SUBSTRAIT_TYPE(I64),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp32),
+                                                  CREATE_SUBSTRAIT_TYPE(Fp64)},
+                                                 {7, 7, 7, 7}))};
+  }
+};
+
+class CiderProjectAllTestBase : public CiderTestBase {
+ public:
+  CiderProjectAllTestBase() {
+    table_name_ = "test";
+    create_ddl_ =
+        "CREATE TABLE test(col_1 INTEGER, col_2 BIGINT, col_3 TINYINT, col_4 SMALLINT, "
+        "col_5 FLOAT, col_6 DOUBLE, col_7 DATE, col_8 BOOLEAN);";
+    input_ = {std::make_shared<CiderBatch>(QueryDataGenerator::generateBatchByTypes(
+        100,
+        {"col_1", "col_2", "col_3", "col_4", "col_5", "col_6", "col_7", "col_8"},
+        {CREATE_SUBSTRAIT_TYPE(I32),
+         CREATE_SUBSTRAIT_TYPE(I64),
+         CREATE_SUBSTRAIT_TYPE(I8),
+         CREATE_SUBSTRAIT_TYPE(I16),
+         CREATE_SUBSTRAIT_TYPE(Fp32),
+         CREATE_SUBSTRAIT_TYPE(Fp64),
+         CREATE_SUBSTRAIT_TYPE(Date),
+         CREATE_SUBSTRAIT_TYPE(Bool)},
+        {1, 2, 2, 2, 3, 3, 4, 2},
+        GeneratePattern::Random))};
+  }
+};
+
+TEST_F(CiderProjectAllTestBase, filterProjectAllTest) {
+  assertQuery("SELECT * FROM test");
+
+  assertQuery("SELECT * FROM test where TRUE");
+
+  assertQuery(
+      "SELECT * FROM test where (col_3 > 0 and col_4 > 0) or (col_5 < 0 and col_6 < 0)");
+
+  assertQuery("SELECT * FROM test where col_2 <> 0 and col_7 > '1972-02-01'");
+
+  assertQuery("SELECT *, 3 >= 2 FROM test where col_8 = true");
+
+  assertQuery(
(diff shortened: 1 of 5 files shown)


Deegue:
This PR should be moved on(review & retest) after https://github.com/intel/BDTK/pull/90 is merged.

cider/tests/functionality/CiderFilterTest.cpp
```diff
@@ -20,265 +20,157 @@
...
+         CREATE_SUBSTRAIT_TYPE(Fp64)},
+        {1, 2, 2, 2, 3, 3},
+        GeneratePattern::Random);
+    input_ = {std::make_shared<CiderBatch>(
+        *(new CiderBatch(schema, array, std::make_shared<CiderDefaultAllocator>())))};
   }
 };
 
 TEST_F(CiderProjectAllTestBase, filterProjectAllTest) {
-  assertQuery("SELECT * FROM test");
```
yma11, line 139: Test cases for old data format need be kept until we finished migration.

cider/tests/functionality/CiderFilterTest.cpp
```diff
@@ -20,265 +20,157 @@
...
-         CREATE_SUBSTRAIT_TYPE(Fp64),
-         CREATE_SUBSTRAIT_TYPE(Date),
-         CREATE_SUBSTRAIT_TYPE(Bool)},
-        {1, 2, 2, 2, 3, 3, 4, 2},
-        GeneratePattern::Random))};
+         CREATE_SUBSTRAIT_TYPE(Fp64)},
+        {1, 2, 2, 2, 3, 3},
+        GeneratePattern::Random);
+    input_ = {std::make_shared<CiderBatch>(
+        *(new CiderBatch(schema, array, std::make_shared<CiderDefaultAllocator>())))};
```
yma11, line 152: The usage seems weird. Can we provide a similar API by doing this things inside `CiderTestBase`?

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -532,7 +532,25 @@ DuckDbResultConvertor::fetchDataToArrowFormattedCiderBatch(
...
+        arrow_array->release = CiderBatchUtils::ciderEmptyArrowArrayReleaser;
+
+        arrow_schema->format = "+s";
+        arrow_schema->dictionary = nullptr;
+        arrow_schema->n_children = 0;
+        arrow_schema->children = (ArrowSchema**)std::malloc(0);
+        arrow_schema->release = CiderBatchUtils::ciderEmptyArrowSchemaReleaser;
+
+        batch_res.push_back(std::make_shared<CiderBatch>(*(new CiderBatch(
+            arrow_schema, arrow_array, std::make_shared<CiderDefaultAllocator>()))));
```
yma11, line 553: I remember DuckDb result convertor should be already implemented by this [PR](https://github.com/intel/BDTK/pull/90). please check.

Deegue: Here we should handle empty batch get from DuckDB. If we return a CiderBatch leaving schema and array nullptr, many methods (like getLength etc.) will core. And I think the PR should be https://github.com/intel/BDTK/pull/67.

cider/tests/functionality/CiderFilterTest.cpp
Deegue: Moved to `CiderTestBase`.

cider/tests/functionality/CiderFilterTest.cpp
Deegue: Reverted and keep all old test cases.

cider/tests/utils/DuckDbQueryRunner.cpp
yma11: Got U. Just make sure there is no duplicated change about this part.

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -532,7 +532,25 @@ DuckDbResultConvertor::fetchDataToArrowFormattedCiderBatch(
     auto chunk = result->Fetch();
     if (!chunk || (chunk->size() == 0)) {
       if (batch_res.empty()) {
-        batch_res.push_back(std::make_shared<CiderBatch>());
+        auto arrow_array = CiderBatchUtils::allocateArrowArray();
+        auto arrow_schema = CiderBatchUtils::allocateArrowSchema();
+
+        arrow_array->length = 0;
```
jikunshang, line 538: these are same to `CiderBatchUtils::allocateArrowArray`?

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -532,7 +532,25 @@ DuckDbResultConvertor::fetchDataToArrowFormattedCiderBatch(
...
+        arrow_array->n_children = 0;
+        arrow_array->buffers = nullptr;
+        arrow_array->n_buffers = 0;
+        arrow_array->private_data = nullptr;
+        arrow_array->children = (ArrowArray**)std::malloc(0);
```
jikunshang, line 543: what does malloc(0) returns ? and we should avoid call malloc directly.

cider/tests/utils/CiderTestBase.h
```diff
@@ -65,6 +73,8 @@ class CiderTestBase : public testing::Test {
   std::string table_name_;
   std::string create_ddl_;
   std::vector<std::shared_ptr<CiderBatch>> input_;
+  ArrowArray* array = nullptr;
```
jikunshang, line 76: `array_` & `schema_`

cider/tests/utils/CiderTestBase.cpp
```diff
@@ -50,12 +50,26 @@ void CiderTestBase::prepareArrowBatch() {
...
                                      const std::string& json_file) {
   auto duck_res = duckDbQueryRunner_.runSql(sql);
-  auto duck_res_batch = DuckDbResultConvertor::fetchDataToCiderBatch(duck_res);
+  auto duck_res_batch =
+      DuckDbResultConvertor::fetchDataToArrowFormattedCiderBatch(duck_res);
+
+  auto cider_input = json_file.size() ? json_file : sql;
+  auto cider_res_batch = std::make_shared<CiderBatch>(
+      ciderQueryRunner_.runQueryOneBatch(cider_input, input_[0], true));
+  EXPECT_TRUE(CiderBatchChecker::checkArrowEq(duck_res_batch, cider_res_batch, false));
```
jikunshang, line 59: so we can remove `checkArrowEqTemp`? and `convertToArrowRepresentation`

cider/tests/utils/CiderTestBase.cpp
```diff
@@ -50,12 +50,26 @@ void CiderTestBase::prepareArrowBatch() {
...
+  auto duck_res = duckDbQueryRunner_.runSql(sql);
+  auto duck_res_batch =
+      DuckDbResultConvertor::fetchDataToArrowFormattedCiderBatch(duck_res);
 
   auto cider_input = json_file.size() ? json_file : sql;
   auto cider_res_batch = std::make_shared<CiderBatch>(
       ciderQueryRunner_.runQueryOneBatch(cider_input, input_[0], true));
-  EXPECT_TRUE(CiderBatchChecker::checkArrowEqTemp(duck_res_batch, cider_res_batch));
+
+  EXPECT_TRUE(CiderBatchChecker::checkEq(duck_res_batch, cider_res_batch, true));
```
YBRua, line 72: I think it probably should be `EXPECT_TRUE(CiderBatchChecker::checkArrowEq(...))` here?

Deegue: Yes, just the same as above, let me revise it.

cider/tests/utils/CiderTestBase.h
haiweizh, line 76: protected variable's name should been end with _

haiweizh (changes requested):

cider/tests/utils/CiderTestBase.h
```diff
@@ -34,7 +34,13 @@
 class CiderTestBase : public testing::Test {
  public:
   void SetUp() override {
-    duckDbQueryRunner_.createTableAndInsertData(table_name_, create_ddl_, input_);
+    if (array && schema) {
+      input_ = {std::make_shared<CiderBatch>(
```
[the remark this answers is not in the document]
Deegue: It works well in local testing. Will it be any trouble if we set input_ here?

cider/tests/utils/DuckDbQueryRunner.cpp
Deegue: redundant code removed.

cider/tests/utils/DuckDbQueryRunner.cpp
Deegue: I think we will get a valid pointer by malloc(0) instead of nullptr. 

Went through all the places using `getChildrenArrayPtr()` and it looks okay if we just set it to nullptr.

cider/tests/utils/CiderTestBase.h
Deegue: done

cider/tests/utils/CiderTestBase.cpp
Deegue: yes, I think so

cider/tests/utils/CiderTestBase.h
Deegue: Get it, thanks~

cider/tests/utils/CiderTestBase.h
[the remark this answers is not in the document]
jikunshang: `auto ptr = std::make_shared<Type>(* new Type())`  
`auto ptr = std::make_shared<Type>()`
the later one is better, right?

cider/tests/utils/CiderTestBase.cpp
jikunshang: seems some types are not totally ready(like string). let's keep them for a while.

cider/tests/functionality/CiderFilterTest.cpp
```diff
@@ -276,9 +261,245 @@ TEST_F(CiderFilterSequenceTestBase, multiColEqualTest) {
...
+      true);
+}
+
+// Extends CiderTestBase and create a (99 rows, 4 types columns) table for filter test.
+class CiderFilterSequenceTestArrow : public CiderTestBase {
+ public:
+  CiderFilterSequenceTestArrow() {
+    table_name_ = "test";
+    create_ddl_ =
+        R"(CREATE TABLE test(col_1 INTEGER, col_2 BIGINT,
```
jikunshang, line 371: Please add `NOT NULL` in create table command if you didn't generate null data. generated IR will be different.

cider/tests/functionality/CiderFilterTest.cpp
```diff
@@ -398,117 +611,135 @@ TEST_F(CiderFilterSequenceTestArrow, ArrowMultiFilterWithAndTest) {
...
 }
 
-TEST_F(CiderFilterRandomTestBase, multiColRandomTest) {
-  assertQuery("SELECT col_1, col_5 FROM test WHERE col_1 < col_5");
-  assertQuery("SELECT col_2, col_6 FROM test WHERE col_2 < col_6");
-  assertQuery("SELECT col_3, col_7 FROM test WHERE col_3 <= col_7");
-  assertQuery("SELECT col_4, col_8 FROM test WHERE col_4 <= col_8");
-  assertQuery("SELECT col_1, col_5 FROM test WHERE col_1 <> col_5");
-  assertQuery(
+TEST_F(CiderFilterRandomTestArrow, multiColRandomTest) {
```
jikunshang, line 620: should we keep original test cases?

Deegue: yes, all original test cases are on top of this file.

cider/tests/functionality/CiderFilterTest.cpp
Deegue: done

cider/tests/functionality/CiderFilterTest.cpp
jikunshang: oh sorry. I didn't notice that. Can we re-org to reduce code diff?

Deegue: never mind, original test cases interspersing with the new ones will reduce code diff, but we have to go into each case to find and remove the original ones when finished arrow migration.

jikunshang (approved):

yma11 (approved):

PR #94, opened 2022-10-24
[POAE7-2465]Support BOOL for arrow format

imajiayu:
### What changes were proposed in this pull request?
Support BOOL for arrow format, add related function in ArrowArrayBuilder and QueryArrowArrayGenerator

### How was this patch tested?
add Unit Test in corresponding test files

PR #101, opened 2022-10-25
[POAE7-2469] arrow bool support for DuckDbRunner and CiderBatchChecker

YBRua:

Added arrow format boolean value support for `DuckDbQueryRunner` and `CiderBatchChecker`.

- `DuckDbQueryRunner` now supports inserting and exporting arrow-format boolean data.
- `CiderBatchChecker` now supports checking if two boolean Arrow array are equal.

Several notes here:

- Boolean values in `ScalarBatch<bool>` are no longer represented by `uint8_t`s. Instead they are now stored as bit-packed values. Correspondingly, `getRawData()` returns a `uint8_t*` pointer that points to the array of bit-packed booleans.
- Test cases in this PR depend on #94. Current UTs are hand-crafted and will be updated after PR94 is merged. Comments and reviews are welcome, but approval of this PR can be deferred until test cases are fully updated.

Migrating unit test framework to Apache Arrow data format.

No

Handcrafted UTs. To be updated after builder PR is merged.


cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -144,6 +144,47 @@ bool CiderBatchChecker::checkOneScalarBatchEqual(const ScalarBatch<T>* expected_
   return true;
 }
 
+template <>
+bool CiderBatchChecker::checkOneScalarBatchEqual<bool>(
```
YBRua, line 148: Not quite sure about the implementation choice here. The check is based on bitwise ops instead of `memcmp`. My intention is to use validity bitmap as a mask to avoid failure cases caused by different padding or null values. But perhaps it can be done in a more efficient way, or maybe I should have just used `memcmp`.

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -144,6 +144,47 @@ bool CiderBatchChecker::checkOneScalarBatchEqual(const ScalarBatch<T>* expected_
...
+    uint8_t expected_masked = expected_validity_buffer
+                                  ? expected_data_buffer[i] & expected_validity_buffer[i]
+                                  : expected_data_buffer[i];
+    uint8_t actual_masked = actual_validity_buffer
+                                ? actual_data_buffer[i] & actual_validity_buffer[i]
+                                : actual_data_buffer[i];
+
+    // for the last byte, clear trailing padding values to prevent false positives caused
+    // by differences in padding
+    if (i == bytes - 1) {
```
jikunshang, line 172: can we move this if block out of for loop?

jikunshang (approved):

cider/tests/utils/CiderBatchChecker.cpp
yma11: Maybe you can use memcmp but with only compare the correct length of bits.

cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -144,6 +144,47 @@ bool CiderBatchChecker::checkOneScalarBatchEqual(const ScalarBatch<T>* expected_
...
+  for (int i = 0; i < bytes; ++i) {
+    // apply bitwise AND
+    uint8_t expected_masked = expected_validity_buffer
+                                  ? expected_data_buffer[i] & expected_validity_buffer[i]
+                                  : expected_data_buffer[i];
+    uint8_t actual_masked = actual_validity_buffer
+                                ? actual_data_buffer[i] & actual_validity_buffer[i]
+                                : actual_data_buffer[i];
+
+    // for the last byte, clear trailing padding values to prevent false positives caused
```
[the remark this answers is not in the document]
YBRua: I think it should be **trailing**, like "trailing whitespaces", meaning "following behind".

cider/tests/utils/CiderBatchChecker.cpp
YBRua: Not sure if I got it right. This `if` block is like a preprocessing step for the last byte. If we moved this `if` out of the `for` loop, we would have to move the check for the last byte out of this loop.

cider/include/cider/batch/ScalarBatch.h
```diff
@@ -85,4 +85,59 @@ class ScalarBatch final : public CiderBatch {
   }
 };
 
+template <>
+class ScalarBatch<bool> final : public CiderBatch {
```
jikunshang, line 89: will this class be conflict to this change? https://github.com/intel/BDTK/pull/104/files#diff-2e23a8d5af6d6d4a8668b9f372249cb3bfb8d450c124df95964ee50110d6c42aR39

jikunshang (dismissed):

cider/include/cider/batch/ScalarBatch.h
YBRua: Oh, didn't notice the change in #104. Seems this one can be removed.

yma11 (approved):

jikunshang (approved):

PR #115, opened 2022-10-31
[POAE7-2509] UT framework support for 128-bit DECIMALs

YBRua:

Updated `DuckDbQueryRunner` and `CiderBatchChecker` to support 128-bit decimal types exported by DuckDb, including `HUGEINT` and `DECIMAL(a, b)` types (both stored as 128-bit integers in Arrow Array).

Both types are stored as 128-bit (16 byte) binaries in the exported array. For `HUGEINT`, we interpret it as an `int128` value, and compare to the `int64 ` value in Cider (we assume no overflow will occur in `int64`). For `DECIMAL`, we convert it to `double` and compare with the `double`values in Cider. Both comparisons are based on `ConcatenatedRow`s because direct `memcmp` will always fail as the size of arrow arrays are different.

NOTE: Current implementation is based on GCC's support for `__int128_t` types, and it may require GCC/Clang compilers to work properly.

For supporting UTs that involved `HUGEINT` or `DECIMAL` results.

- `HUGEINT` is typically a result of summation over integer types (e.g. `SELECT SUM(int_col)`).
- `DECIMAL` is typically a result of ops between integer and decimals (e.g. `SELECT int_col + 0.123`).

No.

Unit tests.

### DIFF
--- a/cider/tests/utils/CiderBatchChecker.cpp
+++ b/cider/tests/utils/CiderBatchChecker.cpp
@@ -251,6 +251,10 @@ bool CiderBatchChecker::checkOneStructBatchEqual(CiderBatch* expected_batch,
             checkOneScalarBatchEqual<double>(expected_child->as<ScalarBatch<double>>(),
                                              actual_child->as<ScalarBatch<double>>());
         break;
+      case SQLTypes::kDECIMAL:
+        // duckdb decimals use 16 bytes, but cider decimals use 8 bytes
+        // therefore buffer compare will never pass, no need to check it
+        return false;
       case SQLTypes::kSTRUCT:
         is_equal = checkOneStructBatchEqual(expected_child.get(), actual_child.get());
         break;
--- a/cider/tests/utils/CiderBatchStringifier.cpp
+++ b/cider/tests/utils/CiderBatchStringifier.cpp
@@ -95,13 +95,13 @@ std::string StructBatchStringifier::stringifyValueAt(CiderBatch* batch, int row_
   return row.getString();
 }
 
-uint8_t DecimalBatchStringifier::getScale(const ScalarBatch<CiderInt128>* batch) {
+uint8_t DecimalBatchStringifier::getScale(const ScalarBatch<__int128_t>* batch) {
   auto type_str = std::string(batch->getArrowFormatString());
   uint8_t scale = std::stoi(type_str.substr(type_str.find(',') + 1));
   return scale;
 }
 
-uint8_t DecimalBatchStringifier::getWidth(const ScalarBatch<CiderInt128>* batch) {
+uint8_t DecimalBatchStringifier::getWidth(const ScalarBatch<__int128_t>* batch) {
   auto type_str = std::string(batch->getArrowFormatString());
   auto start = type_str.find(':') + 1;
   auto end = type_str.find(',');
@@ -110,7 +110,7 @@ uint8_t DecimalBatchStringifier::getWidth(const ScalarBatch<CiderInt128>* batch)
 }
 
 std::string DecimalBatchStringifier::stringifyValueAt(CiderBatch* batch, int row_index) {
-  auto scalar_batch = batch->as<ScalarBatch<CiderInt128>>();
+  auto scalar_batch = batch->as<ScalarBatch<__int128_t>>();
   if (!scalar_batch) {
     CIDER_THROW(CiderRuntimeException,
                 "ScalarBatch is nullptr, maybe check your casting?");
@@ -124,7 +124,7 @@ std::string DecimalBatchStringifier::stringifyValueAt(CiderBatch* batch, int row
   if (valid_bitmap && !CiderBitUtils::isBitSetAt(valid_bitmap, row_index)) {
     return NULL_VALUE;
   } else {
-    CiderInt128 value = data_buffer[row_index];
+    __int128_t value = data_buffer[row_index];
 
     if (!scale) {
       // integral type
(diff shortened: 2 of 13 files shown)


cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -251,6 +251,10 @@ bool CiderBatchChecker::checkOneStructBatchEqual(CiderBatch* expected_batch,
             checkOneScalarBatchEqual<double>(expected_child->as<ScalarBatch<double>>(),
                                              actual_child->as<ScalarBatch<double>>());
         break;
+      case SQLTypes::kDECIMAL:
+        // duckdb decimals use 16 bytes, but cider decimals use 8 bytes
```
jikunshang, line 255: maybe in some cases we can compare ScalarBatch<int64_t/int32_t> and ScalarBatch<__int128_t>

cider/tests/utils/CiderInt128.h
```diff
@@ -0,0 +1,38 @@
...
+#include <cstdint>
+#include <limits>
+#include <string>
+
+#include "cider/CiderBatch.h"
+
+class CiderInt128Utils {
+ public:
+  static std::string Int128ToString(__int128_t input);
+  static std::string Decimal128ToString(__int128_t input, uint8_t width, uint8_t scale);
```
jikunshang, line 34: better name to scale/precision?  https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow14Decimal128TypeE

cider/tests/utils/CiderBatchStringifier.cpp
```diff
@@ -92,6 +95,51 @@ std::string StructBatchStringifier::stringifyValueAt(CiderBatch* batch, int row_
...
+
+  if (valid_bitmap && !CiderBitUtils::isBitSetAt(valid_bitmap, row_index)) {
+    return NULL_VALUE;
+  } else {
+    __int128_t value = data_buffer[row_index];
+
+    if (!scale) {
+      // integral type
+      return CiderInt128Utils::Int128ToString(value);
+    } else {
```
jikunshang, line 132: why not call Decimal128ToString?

cider/tests/utils/DuckDbQueryRunner.cpp
```diff
@@ -574,8 +574,8 @@ CiderBatch DuckDbResultConvertor::fetchOneBatch(
       case ::duckdb::LogicalTypeId::FLOAT:
         addColumnDataToCiderBatch<float>(chunk, i, row_num, col_buf);
         break;
-      case ::duckdb::LogicalTypeId::DOUBLE:
       case ::duckdb::LogicalTypeId::DECIMAL:
+      case ::duckdb::LogicalTypeId::DOUBLE:
```
jikunshang, line 578: unnecessary change? please revert back.

cider/tests/utils/CiderBatchChecker.cpp
YBRua: I don't think so, if "compare" refers to comparing with `memcmp`. For `int64_t`, its buffer uses 8 bytes for each value, for `int128_t`, the buffer uses 16 bytes. The buffer sizes are not aligned so a direct `memcmp` is not feasible. But running a for-loop and compare the data one by one is possible.

cider/tests/utils/CiderBatchStringifier.cpp
```diff
@@ -92,6 +95,51 @@ std::string StructBatchStringifier::stringifyValueAt(CiderBatch* batch, int row_
   return row.getString();
 }
 
+uint8_t DecimalBatchStringifier::getScale(const ScalarBatch<__int128_t>* batch) {
+  auto type_str = std::string(batch->getArrowFormatString());
+  uint8_t scale = std::stoi(type_str.substr(type_str.find(',') + 1));
+  return scale;
+}
+
+uint8_t DecimalBatchStringifier::getWidth(const ScalarBatch<__int128_t>* batch) {
```
yma11, line 104: Does `width` mean `precision`? if so, better use `getPrecision()`.

cider/tests/utils/CiderBatchStringifier.cpp
YBRua: Mainly for consistency with stringifying `double`s. `double`s are converted to strings with 16 significant digits, but the number of significant digits of Decimal128ToString is determined by `scale`. So we first convert a decimal to `double`, and then stringify it to ensure consistency with the results in Cider (which are stored as `double`s)

cider/tests/utils/CiderBatchStringifier.cpp
```diff
@@ -92,6 +95,51 @@ std::string StructBatchStringifier::stringifyValueAt(CiderBatch* batch, int row_
...
+  auto valid_bitmap = scalar_batch->getNulls();
+
+  if (valid_bitmap && !CiderBitUtils::isBitSetAt(valid_bitmap, row_index)) {
+    return NULL_VALUE;
+  } else {
+    __int128_t value = data_buffer[row_index];
+
+    if (!scale) {
+      // integral type
+      return CiderInt128Utils::Int128ToString(value);
```
yma11, line 131: I am wondering for this `scale==0` case, it also can be covered by `CiderInt128Utils::Decimal128ToDouble(value, width, scale)`.

YBRua: The stringifier treats integral types and floating point types differently. Integral types are directly stringified with `to_string`, while floating points are stringified with 16 significant digits.

`scale == 0` means the value being stringified here is an `int128` instance. Its value is expected to be checked with another reference integer (which typically comes from Cider), and in this case we need not (and should not) set any significant digit. Otherwise, the value is a decimal, which will be casted to double and compare with float/double values.

jikunshang (approved):

yma11 (approved):

PR #117, opened 2022-10-31
[POAE7-2484] Support postJoinFilter in Substrait2EU module

haiweizh:

Support postJoinFilter in Substrait2EU module.

Velox's substrait plan contains postJoinFilter in Join Node, which is not support in Substrait2EU module, update that to deal postJoinFilter condition.

No.

UTs.

cider,test


haiweizh:
SQL Such as "SELECT * FROM l JOIN r ON l_a = r_a WHERE l_a < 10", the substrait-json plan usually is JoinRel inside and FilterRel outside. But there is another situation, the filter condition is in postJoinFilter which is inside JoinRel.
So we update Substrait2EU module to support postJoinFilter of JoinRel. 
Plz help to take a review. @jikunshang @yma11 @chaojun-zhang

yma11 (approved):

yma11:
Overall LGTM. I see you put all post-join filters into `join_quals` but not sure it's always valid. Let's keep eyes on this part.

cider/tests/substrait_plan_files/post_join_filter2.json
```diff
@@ -0,0 +1,363 @@
...
+                          }
+                        },
+                        "rootReference": {
+                        }
+                      }
+                    }
+                  }]
+                }
+              },
+              "postJoinFilter": {
```
spevenhe, line 173: So, the ```where``` or ```and``` in sql will be translated to ```postJoinFilter``` in substrait plan by upstream caller?

haiweizh: Yes, part of upstream caller will. This situation is encountered during the join integration of velox. The substrait plan transmitted from velox will put the filter in joinrel's postJoinFilter.

haiweizh: And the filter condition which is putted into Join quals will make the computing faster.

chaojun-zhang: if the join type is semi or anti , can we push down the postJoinFilter?

haiweizh: Substrate2EU supports various types of joins when parsing joins.

spevenhe (approved):
LGTM

PR #120, opened 2022-10-31
[POAE7-2425] add error-check flow for Arith codegen with Arrow and verify proj op

yma11:
## What changes were proposed in this pull request?
add error-check flow for Arith codegen with Arrow and verify proj op

make sure project related codegen work with correct error-handling

No

UT


cider/exec/template/ArithmeticIR.cpp
```diff
@@ -117,61 +119,105 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenArithFun(
...
-                  ? cgen_state_->ir_builder_.CreateAdd(lh_value, rh_value)
-                  : cgen_state_->ir_builder_.CreateFAdd(lh_value, rh_value);
-      break;
-    case kMULTIPLY:
-      value = lh_value->getType()->isIntegerTy()
-                  ? cgen_state_->ir_builder_.CreateMul(lh_value, rh_value)
-                  : cgen_state_->ir_builder_.CreateFMul(lh_value, rh_value);
-      break;
+    case kMULTIPLY: {
+      // add overflow check for only INT-like types?
```
kd97, line 185: https://releases.llvm.org/2.6/docs/LangRef.html#int_overflow. LLVM only provide integer overflow check, I think this is because  floating-point return result will become an infinity rather than wrapping around like integer.

cider/exec/template/ArithmeticIR.cpp
```diff
@@ -850,6 +917,105 @@ llvm::Function* CodeGenerator::getArithWithOverflowIntrinsic(
...
+    const SQLTypeInfo& ti) {
+  AUTOMATIC_IR_METADATA(cgen_state_);
+  auto lhs_lv = lhs->getValue(), rhs_lv = rhs->getValue();
+  auto lhs_null = lhs->getNull(), rhs_null = rhs->getNull();
+  llvm::BasicBlock* div_ok{nullptr};
+  llvm::BasicBlock* div_zero{nullptr};
+  div_ok = llvm::BasicBlock::Create(
+      cgen_state_->context_, "div_ok", cgen_state_->current_func_);
+  if (!null_check_suffix.empty()) {
+    codegenSkipOverflowCheckForNullForArrow(lhs_null, rhs_null, div_ok, ti);
```
kd97, line 934: better name codegenSkipCheckForNullForArrow?

yma11: This name is changed based on origin method name and the method only contains overflow check inside which makes this name more suitable.

jikunshang (approved):

kd97 (dismissed):

kd97 (approved):

jikunshang (approved):

spevenhe (approved):

PR #122, opened 2022-11-01
[POAE7-2479] Use CiderBatch as data type for join bridge to bypass existing Velox memory ownership mechnism

yeyuqiang:
Make CiderJoinBridge to deliver CiderBatch.

Upstream op will destroy CiderBatch after it closed which may cause memory issue for downstream op.

No

Existing UT. 

veloxIntegration

### DIFF
--- a/cider-velox/src/CiderJoinBuild.cpp
+++ b/cider-velox/src/CiderJoinBuild.cpp
@@ -25,7 +25,7 @@
 
 namespace facebook::velox::plugin {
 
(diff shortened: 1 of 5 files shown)


winningsix:
Any test we can add to cover this bug?

yeyuqiang:
> Any test we can add to cover this bug?

Add a project node, the workflow will be project -> hashjoin. this case will cause memory issue previously

winningsix (approved):

chaojun-zhang (approved):
LGTM

PR #129, opened 2022-11-02
[POAE7-2490] String op support for arrow format

jikunshang:

add string op (like substring) support for arrow format.
1. refactor string hasher
2. like function support.
3. add string op codegen support, add some uts
4. string test leverage new dataGenerator/batchChecker

NO

UT.

### DIFF
--- a/cider/exec/operator/aggregate/CiderAggHashTable.cpp
+++ b/cider/exec/operator/aggregate/CiderAggHashTable.cpp
@@ -172,9 +172,6 @@ uint64_t CiderHasher::updateHashMode(const uint64_t entry_num_limit,
 
 void CiderStringHasher::copyStringToLocal(UStringVal* uString) {
   auto size = uString->size();
-  if (size <= sizeof(int64_t)) {
-    return;
-  }
   if (distinctStringsBytes_ > kMaxDistinctStringsBytes) {
     distinctOverflow_ = true;
     return;
@@ -213,14 +210,14 @@ int64_t CiderStringHasher::lookupIdByValue(CiderByteArray value) {
     if (pair.second) {
       if (uStringVals.size() > kMaxDistinct) {
         distinctOverflow_ = true;
-        CIDER_THROW(CiderCompileException, "Overflow in distinct string hash set.");
+        CIDER_THROW(CiderRuntimeException, "Overflow in distinct string hash set.");
       }
       copyStringToLocal(&*pair.first);
       stringCacheVec.emplace_back(*pair.first);
     }
     return uStringVals.find(uString)->id();
   } else {
-    CIDER_THROW(CiderCompileException, "Overflow in distinct string hash set.");
+    CIDER_THROW(CiderRuntimeException, "Overflow in distinct string hash set.");
   }
   return -1;
 }
--- a/cider/exec/operator/aggregate/CiderAggHashTable.cpp
+++ b/cider/exec/operator/aggregate/CiderAggHashTable.cpp
@@ -223,7 +223,7 @@ int64_t CiderStringHasher::lookupIdByValue(CiderByteArray value) {
 }
 
 const CiderByteArray CiderStringHasher::lookupValueById(int64_t id) const {
-  if (0 == id) {
+  if (0 == id || -1 == id) {
     return {0, nullptr};
   }
   UStringVal uString = stringCacheVec[id - 1];
--- a/cider/exec/operator/aggregate/CiderAggHashTable.cpp
+++ b/cider/exec/operator/aggregate/CiderAggHashTable.cpp
@@ -170,7 +170,7 @@ uint64_t CiderHasher::updateHashMode(const uint64_t entry_num_limit,
   return entry_num_limit_max;
 }
 
-void CiderHasher::copyStringToLocal(UStringVal* uString) {
+void CiderStringHasher::copyStringToLocal(UStringVal* uString) {
   auto size = uString->size();
   if (size <= sizeof(int64_t)) {
     return;
@@ -201,7 +201,7 @@ void CiderHasher::copyStringToLocal(UStringVal* uString) {
   uString->setData(str->data() + start);
 }
 
-int64_t CiderHasher::lookupIdByValue(CiderByteArray value) {
+int64_t CiderStringHasher::lookupIdByValue(CiderByteArray value) {
   if (!value.ptr) {
     return 0;
   }
@@ -225,7 +225,7 @@ int64_t CiderHasher::lookupIdByValue(CiderByteArray value) {
   return -1;
 }
 
-const CiderByteArray CiderHasher::lookupValueById(int64_t id) const {
+const CiderByteArray CiderStringHasher::lookupValueById(int64_t id) const {
   if (0 == id) {
     return {0, nullptr};
   }
--- a/cider/exec/operator/aggregate/CiderAggHashTable.h
+++ b/cider/exec/operator/aggregate/CiderAggHashTable.h
@@ -108,6 +108,7 @@ struct UStringValCmp {
 
 class CiderStringHasher {
  public:
+  // id = 0 => empty string, id = -1 => null
   int64_t lookupIdByValue(CiderByteArray value);
   const CiderByteArray lookupValueById(int64_t id) const;
 
--- a/cider/exec/operator/aggregate/CiderAggHashTable.h
+++ b/cider/exec/operator/aggregate/CiderAggHashTable.h
@@ -112,7 +112,7 @@ class CiderStringHasher {
   const CiderByteArray lookupValueById(int64_t id) const;
 
  private:
-  // Stop counting distinct values after this many and revert to regular hash.
+  // Stop counting distinct values after this threshold and revert to regular hash.
   static constexpr int32_t kMaxDistinct = 100'000;
   static constexpr uint32_t kStringBufferUnitSize = 1024;
   static constexpr uint64_t kMaxDistinctStringsBytes = 1 << 20;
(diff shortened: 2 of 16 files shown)


cider/exec/template/Execute.h
```diff
@@ -669,6 +672,8 @@ class Executor {
   }
 
   std::unique_ptr<CgenState> cgen_state_;
+  CiderStringHasher* stringHasher_;
+  mutable std::mutex str_hasher_mutex_;
```
harborn, line 676: why this std::mutex variable is mutable?

cider/exec/template/StringOpsIR.cpp
```diff
@@ -96,6 +96,55 @@ apply_string_ops_and_encode(const char* str_ptr,
   return string_dict_proxy->getOrAddTransient(result_str);
 }
 
+extern "C" RUNTIME_EXPORT int64_t
+apply_string_ops_and_encode_cider(const char* str_ptr,
+                                  const int32_t str_len,
+                                  const int64_t string_ops_handle,
+                                  const int64_t string_hasher_handle) {
```
harborn, line 103: I think the `const` is not necessary for the three arguments.

cider/tests/functionality/CiderStringTest.cpp
```diff
@@ -255,19 +255,150 @@ class CiderStringTestArrow : public CiderTestBase {
...
+    QueryArrowDataGenerator::generateBatchByTypes(
+        schema_,
+        array_,
+        30,
+        {"col_1", "col_2"},
+        {CREATE_SUBSTRAIT_TYPE(I32), CREATE_SUBSTRAIT_TYPE(Varchar)},
+        {2, 2},
+        GeneratePattern::Sequence,
+        0,
+        10);
```
harborn, line 286: why function call's code style with one argument one line?
I think the following code more better:
```c++
QueryArrowDataGenerator::generateBatchByTypes(
    schema_, array_, 30, {"col_1", "col_2"},
    {CREATE_SUBSTRAIT_TYPE(I32), CREATE_SUBSTRAIT_TYPE(Varchar)},
    {2, 2}, GeneratePattern::Sequence, 0, 10);
```

jikunshang: formatter will do this.

cider/exec/operator/aggregate/CiderAggHashTable.h
```diff
@@ -106,6 +106,36 @@ struct UStringValCmp {
   }
 };
 
+class CiderStringHasher {
```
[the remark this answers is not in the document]
jikunshang: yes.

cider/exec/template/StringOpsIR.cpp
```diff
@@ -96,6 +96,55 @@ apply_string_ops_and_encode(const char* str_ptr,
...
+  const auto result_str = string_ops->operator()(raw_str);
+
+  // add this temp result to cache
+  int64_t id = string_hasher->lookupIdByValue(
+      CiderByteArray(result_str.length(), (const uint8_t*)result_str.c_str()));
+  return id;
+}
+
+extern "C" RUNTIME_EXPORT int64_t
+apply_string_ops_and_encode_cider_nullable(const char* str_ptr,
```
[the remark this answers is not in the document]
jikunshang: see L360

cider/exec/operator/aggregate/CiderAggHashTable.h
```diff
@@ -282,6 +286,8 @@ class CiderAggHashTable {
   const std::vector<size_t>& getTargetIndexMap() const { return target_index_map_; }
   CiderHasher& getHasher() { return hasher_; }
   const CiderHasher& getHasher() const { return hasher_; }
+  CiderStringHasher& getStringHasher() { return stringHasher_; }
```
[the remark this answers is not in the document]
jikunshang: not sure. just keep same to `getHasher`. will have a try to remove.

cider/exec/template/Execute.h
jikunshang: same as line 705.

cider/exec/operator/aggregate/CiderAggHashTable.h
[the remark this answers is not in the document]
jikunshang: For string type, we provide two hash table like API via CiderHasher before, and it have nothing related to other CiderHasher logic, So I refactor it to a new class and just for string usage.

cider/exec/template/StringOpsIR.cpp
jikunshang: yes it's not necessary, but it could give user hints that they should not be modifiable.

cider/exec/template/Execute.cpp
```diff
@@ -334,6 +335,10 @@ StringDictionaryProxy* Executor::getStringDictionaryProxy(
   return row_set_mem_owner->getOrAddStringDictProxy(db_id_, dict_id_in, with_generation);
 }
 
+CiderStringHasher* Executor::getCiderStringHasherHandle() const {
+  return stringHasher_;
+}
+
 StringDictionaryProxy* RowSetMemoryOwner::getOrAddStringDictProxy(
```
yma11, line 342: is it okay to replace `StringDictProxy` with `CiderStringHasher`? For now, I think it won't work for case like `substr(col_str) in ['a', 'b', 'c']`.

jikunshang: substring() in case will not work well now. we may need redesign.

cider/exec/template/StringOpsIR.cpp
```diff
@@ -270,6 +319,66 @@ llvm::Value* CodeGenerator::codegenPerRowStringOper(const Analyzer::StringOper*
...
+std::unique_ptr<CodegenColValues> CodeGenerator::codegenStringOpExpr(
+    const Analyzer::StringOper* expr,
+    const CompilationOptions& co) {
+  AUTOMATIC_IR_METADATA(cgen_state_);
+
+  CHECK_GE(expr->getArity(), 1UL);
+  CHECK(expr->hasNoneEncodedTextArg());
+
+  const auto& expr_ti = expr->get_type_info();
+  const auto primary_arg = remove_cast(expr->getArg(0));
```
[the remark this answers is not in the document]
jikunshang: it may have a UOper cast to dict encoded to be consistent.

cider/exec/template/StringOpsIR.cpp
```diff
@@ -270,6 +319,66 @@ llvm::Value* CodeGenerator::codegenPerRowStringOper(const Analyzer::StringOper*
...
+  if (str_values->getNull()) {
+    func_name.append("_nullable");
+    string_oper_lvs.push_back(str_values->getNull());
+  }
+  auto id = cgen_state_->emitExternalCall(
+      func_name, get_int_type(64, cgen_state_->context_), string_oper_lvs);
+  llvm::Value* res_null = nullptr;
+  if (str_values->getNull()) {
+    res_null = cgen_state_->emitExternalCall(
+        "cider_check_string_id_is_null", get_int_type(1, cgen_state_->context_), {id});
```
yma11, line 368: Why not directly equal to null of original column?

jikunshang: make sense for most cases. I am not sure whether we have case like `str_not_null` -> `string_op` -> `null`, if have we can reserve for it.

yma11: Maybe in above scenario, we should take it as `empty` string instead of `null`?

cider/exec/operator/aggregate/CiderAggHashTable.cpp
```diff
@@ -170,11 +170,8 @@ uint64_t CiderHasher::updateHashMode(const uint64_t entry_num_limit,
   return entry_num_limit_max;
 }
 
-void CiderHasher::copyStringToLocal(UStringVal* uString) {
+void CiderStringHasher::copyStringToLocal(UStringVal* uString) {
   auto size = uString->size();
-  if (size <= sizeof(int64_t)) {
```
Deegue, line 175: I think we should keep it since we don't need to copy a string whose length is less than the size of a pointer to `uStringValStorage_`.

Deegue (approved):
LGTM

cider/exec/operator/aggregate/CiderAggHashTable.cpp
jikunshang: Based on my understanding, if keep this, it will not save a copy in `uStringValStorage_`,  and cannot find the data from `uStringValStorage_` when use(look up) this string via id.

Deegue: When we call `lookupValueById`, it will always find string through `stringCacheVec` no matter the string length is above or below 8 bytes. 

Besides, `uStringValStorage_` is designed to store long strings in a contiguous block of memory.
Ⅰ. When the length of string is less than or equal to 8 bytes (size of a pointer), we store the pointer to `stringCacheVec` directly. The memory to this pointer is allocated in `unordered_flat_set` when doing `insert`.
Ⅱ. When the length of string is more than 8 bytes, we store them into one or more contiguous blocks of memory. New memory will be allocated and pointers will be changed in `copyStringToLocal`.

jikunshang: For less than or equal 8 bytes scenario, the pointer may be freed, right?

yma11 (dismissed):

cider/exec/operator/aggregate/CiderAggHashTable.cpp
Deegue: You are right, inserting into `unordered_flat_set` will change the pointer of `UStringVal` but the pointer of data won't be changed. We should remove this part of code and it seems not efficient for memory in less than or equal 8 bytes scenario.

cider/exec/template/CastIR.cpp
```diff
@@ -54,6 +54,7 @@ llvm::Value* CodeGenerator::codegenCast(const Analyzer::UOper* uoper,
 std::unique_ptr<CodegenColValues> CodeGenerator::codegenCastFun(
     const Analyzer::UOper* uoper,
     const CompilationOptions& co) {
+  std::cout << uoper->toString() << std::endl;
```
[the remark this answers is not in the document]
jikunshang: done.

yma11 (approved):

Deegue (approved):

spevenhe (approved):

PR #130, opened 2022-11-02
[POAE7-2529] fix some codes of macros define

harborn:


cider/function/scalar/RuntimeFunctions.h
```diff
@@ -119,10 +119,10 @@ extern "C" void agg_count_distinct_bitmap(int64_t* agg,
                                           const int64_t val,
                                           const int64_t min_val);
 
-#define EMPTY_KEY_64 std::numeric_limits<int64_t>::max()
-#define EMPTY_KEY_32 std::numeric_limits<int32_t>::max()
-#define EMPTY_KEY_16 std::numeric_limits<int16_t>::max()
-#define EMPTY_KEY_8 std::numeric_limits<int8_t>::max()
+constexpr int64_t empty_key_64 = std::numeric_limits<int64_t>::max();
```
winningsix, line 122: Q: why use lower cased name for ```constexpr```?

winningsix (approved):
Just a minor question, other part LGTM. Good catch!

cider/function/scalar/RuntimeFunctions.h
harborn: those const variable are used in many code files
defined as macro is not a good style, because:
1. macro just codes replace, it does not have type
2. macro's scope is global, easy conflict when have same name macro definition.

chaojun-zhang (approved):
LGTM

PR #131, opened 2022-11-03
[POAE7-2403] [WIP] Enabled Parquet format data in CICD. 

kyotoyx:

Commented queries that not supported in current version. 

To enable parquet format. 

NO

CICD

CICD


harborn (approved):

winningsix:
Sounds like Parquet format will fail a lot of test cases? Any possibility we enable both dwrf format and Parquet format for short term and deprecate dwrf format once Parquet getting mature. Your thoughts?

kyotoyx:
> Sounds like Parquet format will fail a lot of test cases? Any possibility we enable both dwrf format and Parquet format for short term and deprecate dwrf format once Parquet getting mature. This is to keep current test coverage.
> 
> Your thoughts?
Most of the failing cases are caused by Velox Parquet reader. To support those cases we might do some research to find out if these problem can be solved easily. And if not, we might wait an uncertain period time for the upstream supporting. 
So my suggesting is that, we need to do a minor research (0.5~2 days at most) then decide which format to use in our CICD.

winningsix:
Just to confirm, the merge criteria for this PR is to pass all nightly test, right?

jikunshang:
should we close this?

PR #135, opened 2022-11-03
[POAE7-2531] Support Groupby String for Arrow Format

Deegue:
1. Support string type group_key in Arrow Format.
2. Add fixedChar of substrait type.
3. Add ignore order option for assertQueryArrow.

Function support & Improvement.

No.

UTs.

### DIFF
--- a/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
+++ b/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
@@ -129,7 +129,32 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
   }
 
   void extract(const std::vector<const int8_t*>& rowAddrs, CiderBatch* output) override {
-    UNREACHABLE();
+    size_t rowNum = rowAddrs.size();
+    auto varcharOutput = output->asMutable<VarcharBatch>();
+
+    CHECK(varcharOutput->resizeBatch(rowNum));
+    uint8_t* buffer = varcharOutput->getMutableRawData();
+    int32_t* offset = varcharOutput->getMutableRawOffset();
+    uint8_t* nulls = varcharOutput->getMutableNulls();
+
+    offset[0] = 0;
+    int64_t null_count = 0;
+    for (size_t i = 0; i < rowNum; ++i) {
+      const int8_t* rowPtr = rowAddrs[i];
+      const int64_t* id = reinterpret_cast<const int64_t*>(rowPtr + offset_);
+      CiderByteArray value = hasher_->lookupValueById(*id);
+      // TODO(yizhong): null should not be decided by length
+      if (!value.len) {
+        CiderBitUtils::clearBitAt(nulls, i);
+        ++null_count;
+      } else {
+        offset[i + 1] = offset[i] + value.len;
+        varcharOutput->resizeDataBuffer(offset[i + 1]);
+        buffer = varcharOutput->getMutableRawData();
+        std::memcpy(buffer + offset[i], value.ptr, value.len);
+      }
+    }
+    output->setNullCount(null_count);
   }
 
  private:
--- a/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
+++ b/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
@@ -149,7 +149,7 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
         offset[i + 1] = offset[i];
       } else {
         offset[i + 1] = offset[i] + value.len;
-        varcharOutput->resizeDataBufferIfNeeded(offset[i + 1]);
+        varcharOutput->resizeDataBufferIfNeeded(offset[i], offset[i + 1]);
         buffer = varcharOutput->getMutableRawData();
         std::memcpy(buffer + offset[i], value.ptr, value.len);
       }
--- a/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
+++ b/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
@@ -149,7 +149,7 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
         offset[i + 1] = offset[i];
       } else {
         offset[i + 1] = offset[i] + value.len;
-        varcharOutput->resizeDataBuffer(offset[i + 1]);
+        varcharOutput->resizeDataBufferIfNeeded(offset[i + 1]);
         buffer = varcharOutput->getMutableRawData();
         std::memcpy(buffer + offset[i], value.ptr, value.len);
       }
--- a/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
+++ b/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
@@ -143,10 +143,10 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
       const int8_t* rowPtr = rowAddrs[i];
       const int64_t* id = reinterpret_cast<const int64_t*>(rowPtr + offset_);
       CiderByteArray value = hasher_->lookupValueById(*id);
-      // TODO(yizhong): null should not be decided by length
-      if (!value.len) {
+      if (-1 == value.len) {
         CiderBitUtils::clearBitAt(nulls, i);
         ++null_count;
+        offset[i + 1] = offset[i];
       } else {
         offset[i + 1] = offset[i] + value.len;
         varcharOutput->resizeDataBuffer(offset[i + 1]);
--- a/cider/exec/template/IRCodegen.cpp
+++ b/cider/exec/template/IRCodegen.cpp
@@ -1475,7 +1475,8 @@ Executor::GroupColLLVMValue Executor::groupByColumnCodegen(
 std::unique_ptr<FixedSizeColValues> Executor::groupByColumnCodegen(
     Analyzer::Expr* group_by_col,
     const size_t col_width,
-    const CompilationOptions& co) {
+    const CompilationOptions& co,
+    llvm::Argument* groups_buffer) {
   // TODO: refactor code from previous function.
   AUTOMATIC_IR_METADATA(cgen_state_.get());
   CHECK_GE(col_width, sizeof(int32_t));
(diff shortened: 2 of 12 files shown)


cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
```diff
@@ -129,7 +129,32 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
   }
 
   void extract(const std::vector<const int8_t*>& rowAddrs, CiderBatch* output) override {
-    UNREACHABLE();
+    size_t rowNum = rowAddrs.size();
+    auto varcharOutput = output->asMutable<VarcharBatch>();
+
+    CHECK(varcharOutput->resizeBatch(rowNum));
+    uint8_t* buffer = varcharOutput->getMutableRawData();
+    int32_t* offset = varcharOutput->getMutableRawOffset();
```
yma11, line 137: won't the `id` is fixed size like 4 bytes? If so, why do we need offset buffer?

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
```diff
@@ -129,7 +129,32 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
...
+    uint8_t* nulls = varcharOutput->getMutableNulls();
+
+    offset[0] = 0;
+    int64_t null_count = 0;
+    for (size_t i = 0; i < rowNum; ++i) {
+      const int8_t* rowPtr = rowAddrs[i];
+      const int64_t* id = reinterpret_cast<const int64_t*>(rowPtr + offset_);
+      CiderByteArray value = hasher_->lookupValueById(*id);
+      // TODO(yizhong): null should not be decided by length
+      if (!value.len) {
```
yma11, line 147: You can check the `nulls`, right?

yma11, line 147: What will happen with insert an entry with `null` key? May be the hasher id is `-1`?

cider/function/aggregate/GroupByRuntime.cpp
```diff
@@ -67,7 +67,7 @@ extern "C" RUNTIME_EXPORT NEVER_INLINE int64_t* get_group_value_cider(
   return agg_hash_table->getGroupTargetPtr(key);
 }
 
-extern "C" RUNTIME_EXPORT ALWAYS_INLINE int32_t
+extern "C" RUNTIME_EXPORT ALWAYS_INLINE int64_t
```
yma11, line 70: Why do this change?

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
Deegue: Well, there is no case to return hasher id `-1` yet, and `null` will be also regarded as a `UStringVal`. Null case will be handled after we supported `null` in `CiderByteArray`.

cider/tests/utils/CiderTestBase.cpp
```diff
@@ -48,15 +48,17 @@ void CiderTestBase::prepareArrowBatch() {
 }
 
 void CiderTestBase::assertQueryArrow(const std::string& sql,
-                                     const std::string& json_file) {
+                                     const std::string& json_file,
```
yma11, line 51: I remember there is a method called `assertQueryArrowIgnoreOrder`, this one is duplicated.

cider/function/aggregate/GroupByRuntime.cpp
Deegue: It should be `int64_t` since we get that value from `lookupIdByValue`.

cider/tests/utils/CiderTestBase.cpp
Deegue: So I think we also need a method called `assertQueryArrowOrdered` or we could just keep the same as before.

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
yma11: I think you can align @jikunshang about this special semantics.  He defined some in his [PR](https://github.com/intel/BDTK/pull/129/files).

cider/tests/utils/CiderTestBase.cpp
yma11: Previous we use the flag to determine ignore order or not. But it's also okay to use `assertQueryArrow` and `assertQueryArrowIgnoreOrdered`, which means that by default it's ordered compared. At least duplication is not allowed.

Deegue: Changed to `assertQueryArrowIgnoreOrder` and removed.

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
```diff
@@ -129,7 +129,32 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
...
+        CiderBitUtils::clearBitAt(nulls, i);
+        ++null_count;
+      } else {
+        offset[i + 1] = offset[i] + value.len;
+        varcharOutput->resizeDataBuffer(offset[i + 1]);
```
yma11, line 152: Maybe rename the method `resizeDataBuffer` to `resizeDataBufferIfNeeded`? Otherwise I thought it will always do the actual resize.

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
jikunshang: see [here](https://github.com/intel/BDTK/pull/129/files#diff-09df0308139f7a448df0a49f37bcc36ed82236515d665834cef9cb14f627fc28R122)

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
```diff
@@ -129,7 +129,32 @@ class SimpleAggExtractor<ST, VarCharPlaceHolder> : public CiderAggTargetColExtra
...
+    offset[0] = 0;
+    int64_t null_count = 0;
+    for (size_t i = 0; i < rowNum; ++i) {
+      const int8_t* rowPtr = rowAddrs[i];
+      const int64_t* id = reinterpret_cast<const int64_t*>(rowPtr + offset_);
+      CiderByteArray value = hasher_->lookupValueById(*id);
+      // TODO(yizhong): null should not be decided by length
+      if (!value.len) {
+        CiderBitUtils::clearBitAt(nulls, i);
+        ++null_count;
```
jikunshang, line 149: offset[i + 1] = offset[i];

cider/exec/template/IRCodegen.cpp
```diff
@@ -1492,6 +1493,13 @@ std::unique_ptr<FixedSizeColValues> Executor::groupByColumnCodegen(
         get_int_type(col_width * 8, cgen_state_->context_)));
 
     return std::unique_ptr<FixedSizeColValues>(fixed_size_group_key);
+  } else if (auto multi_value_group_key =
+                 dynamic_cast<MultipleValueColValues*>(group_key)) {
```
jikunshang, line 1497: should be `TwoValueColValues`

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
Deegue: renamed

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
Deegue: done

cider/exec/template/IRCodegen.cpp
Deegue: done

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
jikunshang: emmm, seems it will always reallocate here.

yma11: There is underlying `memmove ` for each call of this method. It's not acceptable. @Deegue will take a look at this further.

Deegue: Updated.

cider/include/cider/batch/ScalarBatch.h
```diff
@@ -138,6 +138,22 @@ class VarcharBatch final : public CiderBatch {
...
+    CHECK(!isMoved());
+    if (!permitBufferAllocate()) {
+      return false;
+    }
+
+    auto array_holder = reinterpret_cast<CiderArrowArrayBufferHolder*>(getArrayPrivate());
+    size_t capacity = array_holder->getBufferSizeAt(2);
+    if (capacity == 0) {
+      array_holder->allocBuffer(2, 4096);
+    } else if (needed > capacity * 0.9) {  // do reallocate when reach 90% of capacity
```
yma11, line 151: should be `curr + needed` ? which means current size + new request size.

Deegue: Here needed means total needed, it's already current size + new request size.

yma11 (dismissed):

cider/tests/functionality/CiderGroupByTest.cpp
```diff
@@ -20,8 +20,148 @@
...
+                                                  {0, 0, 0, 0},
+                                                  GeneratePattern::Random,
+                                                  1,
+                                                  10);
+  }
+};
+
+TEST_F(CiderGroupByVarcharArrowTest, varcharGroupByTest) {
+  /*single not null group by key*/
+  assertQueryArrowIgnoreOrder("SELECT SUM(col_a), col_c FROM table_test GROUP BY col_c",
```
jikunshang, line 52: do we need the second parameter?

Deegue: removed

jikunshang (approved):

yma11 (approved):

bigPYJ1151:
Emm, this PR not passed UTs but merged, and blocking #142. Any fix patch? @Deegue

yma11:
As shown above, the checks are passed before merge but seems it's too old so fail in new trigger after merge. @Deegue please take a look and we can revert it first if hot fix is not easy.

Deegue:
On it.

Deegue:
Fixed in [[POAE7-2531] [HOTFIX] Fix conflicts with POAE7-2490](https://github.com/intel/BDTK/pull/143), @bigPYJ1151 @yma11 sorry for the inconvenience~

PR #144, opened 2022-11-07
[POAE7-2534] Add string IN and nested substring support for arrow format.

jikunshang:

Add string IN and nested substring support for arrow format.

No

UT


cider/exec/template/CodeGenerator.h
```diff
@@ -175,7 +175,7 @@ class CodeGenerator {
                           const SQLTypeInfo&,
                           const Analyzer::Expr*,
                           const CompilationOptions&);
-
+  // only be called for In expr.
   std::unique_ptr<CodegenColValues> codegenCmpFun(const SQLOps,
```
yma11, line 179: It should be also used by `codegenBinOper`.

cider/exec/template/InValuesIR.cpp
```diff
@@ -126,6 +129,59 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenInValues(
...
+          "look_up_string_id_from_hasher",
+          {lhs_str_ptr, lhs_str_len, cider_string_hasher_handle_lv});
+      return cgen_state_->addInValuesBitmap(in_vals_bitmap)
+          ->codegen(id_val, lhs_null, executor());
+    } else {
+      LOG(INFO) << "Bitmap not created, switch to using OR.";
+    }
+  }
+
+  for (auto in_val : expr->get_value_list()) {
```
yma11, line 176: Here we also need to cover non-string case like integer so can't just call `codegenStringEq`.

cider/exec/template/InValuesIR.cpp
```diff
@@ -226,3 +282,71 @@ std::unique_ptr<InValuesBitmap> CodeGenerator::createInValuesBitmap(
...
+    return nullptr;
+  }
+  const auto stringHasher =
+      ti.is_string() ? executor()->getCiderStringHasherHandle() : nullptr;
+  // FIXME: (yma11) in heavydb, use bitmap only when values count > 3
+  // and switch to normal OR op if not. We make this change because of
+  // type info check will fail for case substring(col, const_1, const_2) = "value"
+  // as substring has SQLTypeInfo(kTEXT, kENCODING_DICT, 0, kNULLT) while
+  // const_1/const_2 has SQLTypeInfo(kVARCHAR, kENCODING_NONE, 0, kNULLT)
+  if (val_count > 3 || ti.is_string()) {
```
yma11, line 304: Since you have provided `codegenStringEq`, will `substring(col, const_1, const_2) = 'abc'` be supported? if so, we can revert the logic to remove `FIXME` here.

cider/exec/template/InValuesIR.cpp
```diff
@@ -126,6 +129,59 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenInValues(
   return std::make_unique<FixedSizeColValues>(result, lhs_null);
 }
 
+std::unique_ptr<CodegenColValues> CodeGenerator::codegenInValuesString(
+    const Analyzer::InValues* expr,
+    const CompilationOptions& co) {
```
yma11, line 134: Here we write a specific `codegen` method for `String` as `TwoValueColValues` always need separate logic and makes some duplicate code. Not sure is there a way to provide other kind of abstract to avoid this.

cider/exec/template/CodeGenerator.h
jikunshang: actually we have two function have same name, one is used in CompareIR.cpp to generate binary cmp op, this one is used in InValuesIR.cpp

cider/exec/template/InValuesIR.cpp
jikunshang: such case is supported now, see [here](https://github.com/intel/BDTK/pull/144/files#diff-02bb724432fbc1bf10058038ebddc0fb5341065824c220fcdb896b6f3cd8a2aeR350) Let me update comments here.

yma11 (approved):

jikunshang:
@winningsix @Deegue @spevenhe Please take a review thanks.

cider/function/aggregate/CiderRuntimeFunctions.h
```diff
@@ -134,6 +135,15 @@ extern "C" ALWAYS_INLINE void cider_agg_id_proj_string_nullable(int8_t* str_data
   }
 }
 
+extern "C" int64_t look_up_string_id_from_hasher(const char* str_ptr,
+                                                 const int32_t str_len,
+                                                 const int64_t string_hasher_handle) {
+  auto string_hasher = reinterpret_cast<CiderStringHasher*>(string_hasher_handle);
+  int64_t id =
+      string_hasher->lookupIdByValue(CiderByteArray(str_len, (const uint8_t*)str_ptr));
```
spevenhe, line 143: If the string is null, would this str_len be some really large random int?

jikunshang: this depends on string extractor, for null string it should return {0, nullptr}, I think it's not necessary to do such check here.

spevenhe (approved):
LGTM

PR #146, opened 2022-11-08
[POAE7-2419] support extract for arrow format

spevenhe:

1. add null check before extract because extract runtime functions are complex and time consuming
2. update tests

support arrow array

No

Uts


spevenhe:
For not null, IR will be 
```%9 = call i64 @extract_month(i64 %8)```


For nullable, IR will be 
``` %11 = call i64 @extract_month(i64 %10, i1 %8) ```

File: cider/exec/template/DateTimeIR.cpp [hunk 1]
```diff
@@ -273,11 +274,12 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenExtract(
...
   }
   const auto extract_fname = get_extract_function_name(extract_expr->get_field());
-
-  llvm::Value* res_lv =
-      cgen_state_->emitExternalCall(extract_fname,
-                                    get_int_type(64, cgen_state_->context_),
-                                    std::vector<llvm::Value*>{fromtime_lv});
+  std::vector<llvm::Value*> extract_lvs{fromtime_lv};
+  if (fromtime_nullable->getNull()) {
+    extract_lvs.push_back(fromtime_nullable->getNull());
```
jikunshang, line 279: I think we can just call a function to check whether `fromtime_nullable->getNull()` is null. Or add else block 
```
else {
extract_lvs.push_back(llvm::Value(True);
}
```

cider/exec/template/DateTimeIR.cpp
```diff
@@ -74,8 +74,36 @@ const char* get_extract_function_name(ExtractField field) {
...
+    case kWEEK:
+    case kWEEK_SUNDAY:
+    case kWEEK_SATURDAY:
+    case kDOY:
+    case kMONTH:
+    case kQUARTER:
+    case kYEAR:
+      return true;
+  }
+  UNREACHABLE();
```
jikunshang, line 101: throw cider exception instead?

spevenhe: During substrait to analyzer translation process, this part is already checked in  https://github.com/intel/BDTK/blob/main/cider/function/datetime/ExtractFromTime.cpp#L301

spevenhe: double thought, add ```cider throw``` here may help future development. updated

cider/exec/template/DateTimeIR.cpp
```diff
@@ -70,12 +70,40 @@ const char* get_extract_function_name(ExtractField field) {
     case kYEAR:
       return "extract_year";
   }
-  UNREACHABLE();
+  CIDER_THROW(CiderUnsupportedException, fmt::format("field is {}", field));
   return "";
 }
 
+bool is_extract_function_complex(ExtractField field) {
```
harborn, line 77: please define the ExtractField's enum with value like this:
```c++
enum ExtractField {
  kYEAR = 0x00000001,
  kQUARTER  = 0x00000002,
  kMONTH = 0x00000004,
  kDAY = 0x00000008,
  kHOUR = 0x00000010,
  kMINUTE = 0x00000020,
  kSECOND = 0x00000040,
  kMILLISECOND = 0x00000080,
  ...
};
```
and for following pattern match, we can use bit operation to make faster

cider/exec/template/DateTimeIR.cpp
```diff
@@ -422,6 +456,33 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenDateTrunc(
...
+// TODO: (spevehe) Deprecate after math runtime mul and div updated
+llvm::Value* CodeGenerator::codegenExtractHighPrecisionTimestampsFun(
+    llvm::Value* ts_lv,
+    const SQLTypeInfo& ti,
+    const ExtractField& field) {
+  AUTOMATIC_IR_METADATA(cgen_state_);
+  CHECK(ti.is_high_precision_timestamp());
+  CHECK(ts_lv->getType()->isIntegerTy(64));
+  if (is_subsecond_extract_field(field)) {
+    const auto result =
```
harborn, line 468: please do not use auto here.

cider/exec/template/DateTimeIR.cpp
```diff
@@ -70,12 +70,40 @@ const char* get_extract_function_name(ExtractField field) {
     case kYEAR:
       return "extract_year";
   }
-  UNREACHABLE();
+  CIDER_THROW(CiderUnsupportedException, fmt::format("field is {}", field));
   return "";
 }
 
+bool is_extract_function_complex(ExtractField field) {
+  switch (field) {
```
harborn, line 78:
```c++
constexpr int field_time = kEPOCH | kDATEEPOCH | kQUARTERDAY | kHOUR | kMINUTE | kSECOND | kMILLISECOND | kMICROSECOND | kNANOSECOND;
constexpr int field_date = kDOW | xxx;
if (field & field_time) return false;
else if (field & field_date) return true;
return false;
```

cider/function/datetime/ExtractFromTime.cpp
```diff
@@ -161,12 +165,20 @@ extern "C" RUNTIME_EXPORT ALWAYS_INLINE int64_t extract_dateepoch(const int64_t
...
+    return 0;
+  }
   int64_t const days_past_epoch = floor_div(timeval, kSecsPerDay);
   return unsigned_mod(days_past_epoch + 3, kDaysPerWeek) + 1;
 }
 
-extern "C" RUNTIME_EXPORT ALWAYS_INLINE int64_t extract_day(const int64_t timeval) {
+extern "C" RUNTIME_EXPORT ALWAYS_INLINE int64_t extract_day(const int64_t timeval,
+                                                            const bool is_null = false) {
+  if (is_null == true) {
```
kd97, line 179: how about using macro to generate both `extract_day`/`extract_day_nullable` methods?

cider/exec/template/DateTimeIR.cpp
```diff
@@ -70,12 +70,27 @@ const char* get_extract_function_name(ExtractField field) {
...
+                                      kWEEK_SATURDAY | kDOY | kMONTH | kQUARTER | kYEAR;
+
+bool is_extract_function_complex(ExtractField field) {
+  if (simple_extract_field & field) {
+    return false;
+  } else if (complex_extract_field & field) {
+    return true;
+  }
+  CIDER_THROW(CiderUnsupportedException, fmt::format("field is {}", field));
+  return false;
```
yma11, line 89: Unnecessary line?

cider/function/datetime/ExtractFromTime.cpp
```diff
@@ -77,7 +77,11 @@ extract_nanosecond(const int64_t lcltime) {
 }
 
 // First day of epoch is Thursday, so + 4 to have Sunday=0.
-extern "C" RUNTIME_EXPORT ALWAYS_INLINE int64_t extract_dow(const int64_t lcltime) {
+extern "C" RUNTIME_EXPORT ALWAYS_INLINE int64_t extract_dow(const int64_t lcltime,
+                                                            const bool is_null = false) {
+  if (is_null == true) {
```
yma11, line 82: `if (is_null)`

yma11: same to others.

yma11 (dismissed):
Overall LGTM.

cider/function/datetime/ExtractFromTime.cpp
spevenhe: in IR this is ```@extract_month(i64 %10, i1 %8)```, so the argument passed to ```is_null``` is in type ```i1``` 1 bit , instead of ```bool```1 byte. So ```if (is_null)``` is not working and using ==true will convert to compare 0 and 1 which will work. Let me put a comment on this.

yma11: ` extract_dow(const int64_t lcltime, const bool is_null = false)`. Seems the arguments here is `bool` type. Can't it be converted implicitly after 1 bit passed in the method?

spevenhe: seems it can't. I tested ```if(is_null)``` and the result is random. Didn't find out the reason why...

cider/exec/template/CodeGenerator.h
```diff
@@ -519,10 +519,15 @@ class CodeGenerator {
                            const bool is_real_str,
                            const CompilationOptions&);
 
+  // Deprecating
```
winningsix, line 522: Minor hint: add ```TODO``` for better remind.

cider/tests/substrait_plan_files/extract/day_of_week_not_null.json
```diff
@@ -0,0 +1,86 @@
+{
```
winningsix, line 1: What's the different from [day_of_week_not_null](https://github.com/intel/BDTK/blob/main/cider/tests/substrait_plan_files/extract/day_of_week.json)? It would be great to add some comments either in this file or [cider/tests/functionality/DateTypeQueryTest.cpp](https://github.com/intel/BDTK/pull/146/files#diff-c6ecb6a2e11833acedb7fb05b892b304d7441a22403c94f477e9482ed5212f77)

winningsix (approved):

jikunshang (approved):

PR #147, opened 2022-11-08
[POAE7-2544] String LOWER/UPPER op support for Arrow format

YBRua:

Updated support for String `LOWER()` and `UPPER()` functions.

1. Enabled existing code for string upper and lower ops. Also added some minor patches to Analyzer and RelAlg EU.
2. Isthmus does not support translating `LOWER` or `UPPER` to Substrait plans. ~~So plan JSON files were created manually according to observations and Substrait docs.~~ JSON query plans are generated with a modified local fork of substrait-java that supports `LOWER` and `UPPER`.
3. Increased nudmber of generated rows of CiderStringTest classes from `30` to `50` to cover lowercase letters in the test.

```asm
; IR SNIPPET of SQL Query: SELECT LOWER(col_2) FROM test;
groupby_nullcheck_true:                           ; preds = %filter_true
  ; extract input strings from arrow array
  %4 = call i8* @cider_ColDecoder_extractArrowBuffersAt(i8* %col_buf1, i64 0)  ; null buffer
  %5 = call i8* @cider_ColDecoder_extractArrowBuffersAt(i8* %col_buf1, i64 1)  ; offset buffer
  %6 = call i8* @cider_ColDecoder_extractArrowBuffersAt(i8* %col_buf1, i64 2)  ; data buffer
  %7 = call i8* @extract_str_ptr_arrow(i8* %6, i8* %5, i64 %pos)
  %8 = call i32 @extract_str_len_arrow(i8* %5, i64 %pos)
  %9 = call i1 @check_bit_vector_clear(i8* %4, i64 %pos)

  ; apply string op function && decode results
  %10 = call i64 @apply_string_ops_and_encode_cider_nullable(
                   i8* %7, i32 %8, i64 93825094087552, i64 93825052641744, i1 %9)
  %11 = call i8* @cider_hasher_decode_str_ptr(i64 %10, i64 93825052641744)
  %12 = call i32 @cider_hasher_decode_str_len(i64 %10, i64 93825052641744)

  ; store results to output buffers
  %13 = getelementptr i64, i64* %group_by_buff, i32 0
  %14 = load i64, i64* %13
  %15 = inttoptr i64 %14 to i8*
  call void @reallocate_string_buffer_if_need(i8* %15, i64 %1)
  %16 = call i8* @cider_ColDecoder_extractArrowBuffersAt(i8* %15, i64 2)
  %17 = call i8* @cider_ColDecoder_extractArrowBuffersAt(i8* %15, i64 1)
  %18 = call i8* @cider_ColDecoder_extractArrowBuffersAt(i8* %15, i64 0)
  call void @cider_agg_id_proj_string_nullable(
              i8* %16, i8* %17, i64 %1, i8* %11, i32 %12, i8* %18, i1 %9)
  br label %filter_false
```

Supporting string ops for Arrow format.

No.

UTs.

### DIFF
--- a/cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
+++ b/cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
@@ -37,7 +37,7 @@ bool getExprUpdatable(std::unordered_map<std::shared_ptr<Analyzer::Expr>, bool> 
 }
 
 bool isStringFunction(std::string function_name) {
-  std::unordered_set<std::string> supportedStrFunctionSet{"substring"};
+  std::unordered_set<std::string> supportedStrFunctionSet{"substring", "lower"};
   return supportedStrFunctionSet.find(function_name) != supportedStrFunctionSet.end();
 }
 
@@ -799,8 +799,8 @@ std::shared_ptr<Analyzer::Expr> Substrait2AnalyzerExprConverter::buildStrExpr(
       function_name.begin(), function_name.end(), function_name.begin(), ::toupper);
   auto string_op_kind = name_to_string_op_kind(function_name);
   switch (string_op_kind) {
-      //    case SqlStringOpKind::LOWER:
-      //      return makeExpr<Analyzer::LowerStringOper>(args);
+    case SqlStringOpKind::LOWER:
+      return makeExpr<Analyzer::LowerStringOper>(args);
       //    case SqlStringOpKind::UPPER:
       //      return makeExpr<Analyzer::UpperStringOper>(args);
       //    case SqlStringOpKind::INITCAP:
--- a/cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
+++ b/cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
@@ -37,7 +37,7 @@ bool getExprUpdatable(std::unordered_map<std::shared_ptr<Analyzer::Expr>, bool> 
 }
 
 bool isStringFunction(std::string function_name) {
-  std::unordered_set<std::string> supportedStrFunctionSet{"substring", "lower"};
+  std::unordered_set<std::string> supportedStrFunctionSet{"substring", "lower", "upper"};
   return supportedStrFunctionSet.find(function_name) != supportedStrFunctionSet.end();
 }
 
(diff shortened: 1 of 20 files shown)


spevenhe (dismissed):
LGTM

yma11:
For the unsupported functions like `lower`, `'upper`, etc, you can submit some PR in [substrait-java](https://github.com/substrait-io/substrait-java) to avoid these json files which need maintain in future.

spevenhe:
> For the unsupported functions like `lower`, `'upper`, etc, can you check whether they are supported in latest [substrait-java](https://github.com/substrait-io/substrait-java)? If not, we can submit a PR.

Seems substrait itself does not support such string functions.
https://github.com/substrait-io/substrait/pull/230

YBRua:
> For the unsupported functions like `lower`, `'upper`, etc, can you check whether they are supported in latest [substrait-java](https://github.com/substrait-io/substrait-java)? If not, we can submit a PR.

Yes. @jikunshang also mentioned this yesterday. I'll take a look.

cider/tests/functionality/CiderStringTest.cpp
```diff
@@ -348,6 +348,35 @@ TEST_F(CiderStringNullableTestArrow, ArrowSubstringTest) {
...
+  ///    <https://substrait.io/relations/logical_relations/#virtual-table>
+  // SELECT LOWER(literal)
+  // assertQueryArrow("SELECT LOWER('ABCDEFG');");
+
+  // SELECT LOWER(literal) FROM table
+  assertQueryArrow("SELECT LOWER('aAbBcCdD12') FROM test;",
+                   "stringop_lower_literal_null.json");
+
+  // SELECT UPPER(column) FROM table
+  assertQueryArrow("SELECT col_2, UPPER(col_2) FROM test;", "stringop_upper_null.json");
```
jikunshang, line 371: I think we can add some case like `SELECT * FROM test WHERE UPPER(col_2) = "AAA" ` once #144 got merged.

cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
```diff
@@ -37,7 +37,7 @@ bool getExprUpdatable(std::unordered_map<std::shared_ptr<Analyzer::Expr>, bool>
 }
 
 bool isStringFunction(std::string function_name) {
-  std::unordered_set<std::string> supportedStrFunctionSet{"substring"};
+  std::unordered_set<std::string> supportedStrFunctionSet{"substring", "lower", "upper"};
```
harborn, line 40: please define the input argument as `const std::string& name`
and cause the supportedStrFunctionSet only has three names, so if we want make this function more fast, codes can be written:
```c++
std::string_view names[] = {"substring", "lower", "upper"};
return std::find(std::cbegin(names), std::cend(names), name) != std::cend(names);
```
or
```c++
std::string_view names[] = {"substring", "lower", "upper"};
for (size_t i = 0; i < sizeof(names); ++i) {
  if (names[i] == name) {
    return true;
  }
return false;
```

cider/tests/functionality/CiderStringTest.cpp
YBRua: added

jikunshang (approved):

yma11 (approved):

PR #150, opened 2022-11-08
[POAE7-2426] verify Agg Ops for Arrow Format

imajiayu:
1. verify **non-groupby** Agg Ops for ArrowFormat except paritialAVG and countDistinct.
2. fix COUNT result is NULL when use arrow as input. 

NO

CiderAggArrowTest.cpp


yma11 (commented):
Please make sure all original TCs included in this new test file and skip the ones which are not enabled yet.

yma11 (approved):

jikunshang (approved):

PR #163, opened 2022-11-11
[ Hotfix ] Correct bit width calculation for logical size considering old and new data format

kd97:

fix tpch failed query.


kd97:
fixed tpch failed query caused by date new format migration, pls take a review. @harborn @kyotoyx

winningsix (approved):

harborn (approved):

PR #167, opened 2022-11-14
[POAE7-2559] Scalar Function String Op: TRIM

YBRua:

Arrow format support for scalar function TRIM (String Op) and its variants (LTRIM/RTRIM).

1. Enabled existing code for TRIM plan nodes to EU conversion.
2. Added patches for TRIM ops.
3. Added tests for TRIM/LTRIM/RTRIM. Substrait plans for TRIM UTs are currently **manually created** because modifying substrait-java to support conversion of TRIM ops is a bit nontrivial.
4. Fixed bug in `Trim::operator()` when dealing with RTRIM. It erroneously retained 1 char in cases where all characters should be trimmed (e.g. `RTRIM('xxxxx', 'x')` returns `'x'` instead of `''`).

Arrow format migration.

No.

UTs.

### DIFF
--- a/cider/function/string/StringOps.cpp
+++ b/cider/function/string/StringOps.cpp
@@ -231,7 +231,7 @@ NullableStrType Trim::operator()(const std::string& str) const {
     // no trimming should be applied
     return str;
   }
-  if (trim_begin == trim_end) {
+  if (trim_begin >= trim_end) {
     // all chars should be trimmed
     return std::string("");
   }
--- a/cider/function/string/StringOps.cpp
+++ b/cider/function/string/StringOps.cpp
@@ -228,7 +228,12 @@ NullableStrType Trim::operator()(const std::string& str) const {
     }
   }
   if (trim_begin == 0 && trim_end == str_len - 1) {
+    // no trimming should be applied
     return str;
+  }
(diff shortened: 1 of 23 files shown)


jikunshang (dismissed):

cider/function/string/StringOps.cpp
```diff
@@ -228,8 +228,13 @@ NullableStrType Trim::operator()(const std::string& str) const {
     }
   }
   if (trim_begin == 0 && trim_end == str_len - 1) {
+    // no trimming should be applied
     return str;
   }
+  if (trim_begin >= trim_end) {
```
spevenhe, line 234: will there be ```>``` condition? It seems only when str_len is 0, trim_end will be less than trim_begin. But this is returned in above code.

YBRua: Yes there will be. Let the query be `SELECT LTRIM('xxxxx', 'x')`. Then `trim_begin` will be `5` and `trim_end` will be `4`, resulting in `trim_begin > trim_end` and `trim_begin > str_len - 1`.

spevenhe (approved):

winningsix (approved):

PR #170, opened 2022-11-14
[POAE7-2457] Support Case/IF Function for Arrow Format with String Function

haiweizh:

Support Case/IF Function for Arrow Format with String Function

Support Case/IF Function for Arrow Format with String Function

No

Uts

CODEREFACTOR,cider,test


cider/exec/template/CaseIR.cpp
```diff
@@ -158,11 +158,14 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenCaseExpr(
   std::unique_ptr<CodegenColValues> case_val =
       codegenCaseExpr(case_expr, case_llvm_type, is_real_str, co);
   if (is_real_str) {
-    // FIXME(haiwei): [POAE7-2457] ,blocking by [POAE7-2415]
-    CIDER_THROW(CiderCompileException,
-                "String type case when is not currently supported.");
+    auto case_expr = dynamic_cast<FixedSizeColValues*>(case_val.get());
```
kd97, line 161: should use `TwoValueColValues` type?

kd97: and may need use new string type encoder and decoder.

spevenhe: Could just use codegenVarCharColVar?

haiweizh: Done.

yma11 (approved):

spevenhe (approved):

PR #171, opened 2022-11-14
[POAE7-2582] support loop join for arrow format

spevenhe:

1. support loop join for arrow format
2. updated loop join codegen for left join type
3. commented out columnvar fetch

support arrow format

No

Uts

### DIFF
--- a/cider/exec/template/IRCodegen.cpp
+++ b/cider/exec/template/IRCodegen.cpp
@@ -717,26 +717,59 @@ std::vector<JoinLoop> Executor::buildJoinLoops(
     bool has_remaining_left_join_quals =
         rem_left_join_quals_it != plan_state_->left_join_non_hashtable_quals_.end() &&
         !rem_left_join_quals_it->second.empty();
-    const auto outer_join_condition_remaining_quals_cb =
-        [this, level_idx, &co](const std::vector<llvm::Value*>& prev_iters) {
-          // when we have multiple quals for the left join in the current join level
-          // we first try to build a hashtable by using one of the possible qual,
-          // and deal with remaining quals as extra join conditions
-          FetchCacheAnchor anchor(cgen_state_.get());
-          addJoinLoopIterator(prev_iters, level_idx + 1);
-          llvm::Value* left_join_cond = cgen_state_->llBool(true);
-          CodeGenerator code_generator(this);
-          auto it = plan_state_->left_join_non_hashtable_quals_.find(level_idx);
-          if (it != plan_state_->left_join_non_hashtable_quals_.end()) {
-            for (auto expr : it->second) {
-              left_join_cond = cgen_state_->ir_builder_.CreateAnd(
-                  left_join_cond,
-                  code_generator.toBool(
-                      code_generator.codegen(expr.get(), true, co).front()));
+
+    std::function<llvm::Value*(const std::vector<llvm::Value*>&)>
+        outer_join_condition_remaining_quals_cb;
+    if (co.use_cider_data_format) {
+      outer_join_condition_remaining_quals_cb =
+          [this, level_idx, &co](const std::vector<llvm::Value*>& prev_iters) {
+            // when we have multiple quals for the left join in the current join level
+            // we first try to build a hashtable by using one of the possible qual,
+            // and deal with remaining quals as extra join conditions
+            FetchCacheAnchor anchor(cgen_state_.get());
+            addJoinLoopIterator(prev_iters, level_idx + 1);
+            llvm::Value* left_join_cond = cgen_state_->llBool(true);
+            CodeGenerator code_generator(this);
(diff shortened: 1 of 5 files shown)


cider/tests/functionality/CiderNonEqualLoopJoinTest.cpp
```diff
@@ -22,164 +22,215 @@
 #include <gtest/gtest.h>
 #include "tests/utils/CiderTestBase.h"
 
-class CiderNonEqualJoinTest : public CiderJoinTestBase {
+// TODO: (spevenhe) For not null query, if set NOT NULL on column ddl, the final project
+// will go to cider_agg_id_proj_xx() instead of cider_agg_id_proj_xx_nullable(), So the
+// returned null values are incorrect.
```
jikunshang, line 27: why results are incorrect?

cider/tests/functionality/CiderNonEqualLoopJoinTest.cpp
```diff
@@ -22,164 +22,215 @@
...
+class CiderArrowSeqNotNullJoinTest : public CiderArrowFormatJoinTestBase {
```
jikunshang, line 28: Please keep vanilla test cases here.

cider/tests/utils/CiderTestBase.cpp
```diff
@@ -179,6 +180,14 @@ void CiderJoinTestBase::assertJoinQueryRowEqualForArrowFormat(
...
 
-  EXPECT_TRUE(CiderBatchChecker::checkArrowEq(
-      duck_res_batches[0], cider_res_batch, ignore_order));
+  if (compare_value) {
+    EXPECT_TRUE(CiderBatchChecker::checkArrowEq(
+        duck_res_batches[0], cider_res_batch, ignore_order));
+  } else {
+    // only check row_num, column_num since duckdb in will return wrong results in left
+    // range join with nulls.
+    EXPECT_TRUE(duck_res_batches[0]->getChildrenNum() ==
```
jikunshang, line 189: EXPECT_EQ

cider/tests/utils/CiderTestBase.cpp
```diff
@@ -168,7 +168,8 @@ void CiderJoinTestBase::assertJoinQueryRowEqual(const std::string& sql,
 void CiderJoinTestBase::assertJoinQueryRowEqualForArrowFormat(
     const std::string& sql,
     const std::string& json_file,
-    const bool ignore_order) {
+    const bool ignore_order,
+    const bool compare_value) {
   VLOG(4) << sql;
```
jikunshang, line 173: remove this?

cider/tests/functionality/CiderNonEqualLoopJoinTest.cpp
spevenhe: Even if we create NOT NULL column, in left join cases, there should be null results when left table failed to find a match in right table. But current columnvar codegen will not go to null handle part if we set column for NOT NULL.

cider/tests/utils/CiderTestBase.cpp
spevenhe: This is debug4 level log to help identify the sql when using ```define``` for test. And in cider codegen part, it also contains many debug level log.

cider/exec/template/IRCodegen.cpp
```diff
@@ -717,26 +717,58 @@ std::vector<JoinLoop> Executor::buildJoinLoops(
...
-            for (auto expr : it->second) {
-              left_join_cond = cgen_state_->ir_builder_.CreateAnd(
-                  left_join_cond,
-                  code_generator.toBool(
-                      code_generator.codegen(expr.get(), true, co).front()));
+    std::function<llvm::Value*(const std::vector<llvm::Value*>&)>
+        outer_join_condition_remaining_quals_cb;
+    if (co.use_cider_data_format) {
+      outer_join_condition_remaining_quals_cb =
+          [this, level_idx, &co](const std::vector<llvm::Value*>& prev_iters) {
```
harborn, line 724: why not implement this lambda as a normal function?
if the lambda convert to a `std::function<llvm::Value*(const std::vector<llvm::Value*>&)>`
the std::function's execution efficiency is really slow.

spevenhe: this part would refactor in JITLib, so does not care about efficiency right now.

cider/tests/functionality/CiderNonEqualLoopJoinTest.cpp
spevenhe: added CiderArrowSeqNotNullInDDLJoinTest

spevenhe: updated. Add original tests

jikunshang (approved):

yma11 (approved):

PR #173, opened 2022-11-15
[POAE7-2594] cast numeric type into string type support

kd97:

support cast primitive type into string

feature support.

No

UTs

### DIFF
--- a/cider/exec/template/CastIR.cpp
+++ b/cider/exec/template/CastIR.cpp
@@ -124,8 +124,19 @@ llvm::Value* CodeGenerator::codegenCastNonStringToString(
     case kDOUBLE:
       fn_call += to_lower(toString(operand_type));
       break;
+    case kTIME:
+      fn_call += "time";
+      break;
+    case kTIMESTAMP:
+      fn_call += "timestamp";
+      operand_lvs.emplace_back(llvm::ConstantInt::get(
+          get_int_type(32, cgen_state_->context_), operand_ti.get_dimension()));
+      break;
+    case kDATE:
+      fn_call += "date";
+      break;
     default:
-      throw std::runtime_error("Unimplemented type for string cast");
+      CIDER_THROW(CiderCompileException, "Unimplemented type for string cast");
   }
   operand_lvs.emplace_back(cider_string_hasher_handle_lv);
   auto str_hashed_lv = cgen_state_->emitExternalCall(
--- a/cider/exec/template/CastIR.cpp
+++ b/cider/exec/template/CastIR.cpp
@@ -69,7 +69,24 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenCastFun(
   auto operand_lv = operand_fixedsize->getValue();
   llvm::Value* target_lv = nullptr;
   if (operand_lv->getType()->isIntegerTy()) {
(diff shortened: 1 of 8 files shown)


cider/exec/template/StringOpsIR.cpp
```diff
@@ -170,7 +169,7 @@ DEF_CONVERT_TO_STRING_AND_ENCODE(double, double)
 extern "C" RUNTIME_EXPORT ALWAYS_INLINE int64_t
 convert_to_string_and_encode_bool(const int8_t operand,
                                   const int64_t string_dict_handle) {
-  std::string str = operand == 1 ? "true" : "false";
+  std::string str = (operand == 1) ? "true" : "false";
```
yma11, line 172: Should it be "1"/"0" or "true"/"false"?

cider/exec/template/CastIR.cpp
```diff
@@ -69,7 +69,24 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenCastFun(
   auto operand_lv = operand_fixedsize->getValue();
   llvm::Value* target_lv = nullptr;
   if (operand_lv->getType()->isIntegerTy()) {
```
jikunshang, line 71: seems will not cast float/double to string?

cider/exec/template/StringOpsIR.cpp
yma11: Offline checked with Yixiu, should be `true` and `false`.

jikunshang (dismissed):

cider/exec/template/StringOpsIR.cpp
```diff
@@ -326,10 +385,8 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenStringOpExpr(
 
   CHECK_GE(expr->getArity(), 1UL);
   CHECK(expr->hasNoneEncodedTextArg());
-
   const auto& return_ti = expr->get_type_info();
-  // Should probably CHECK we have a UOper cast to dict encoded to be consistent
-  const auto primary_arg = remove_cast(expr->getArg(0));
+  const auto primary_arg = expr->getArg(0);
```
spevenhe, line 389: I see in codegenPerRowStringOper, there is still a ```remove_cast(expr->getArg(0));```. Should ```remove_cast``` in there be removed too?

kd97: `codegenPerRowStringOper` old codepath for string op, afraid it will affect old string_dict encoding, so keep it.

yma11 (approved):

cider/exec/template/CastIR.cpp
```diff
@@ -86,6 +103,46 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenCastFun(
...
+    case kTIMESTAMP:
+      fn_call += "timestamp";
+      operand_lvs.emplace_back(llvm::ConstantInt::get(
+          get_int_type(32, cgen_state_->context_), operand_ti.get_dimension()));
+      break;
+    case kDATE:
+      fn_call += "date";
+      break;
+    default:
+      CIDER_THROW(CiderCompileException, "Unimplemented type for string cast");
```
winningsix, line 138: Minor hint: we can mention the exact type for this exception. ```("Unimplemented type {} for string cast", operand_type)```

winningsix (approved):

PR #175, opened 2022-11-16
[POAE7-2596]move codegen to Expr

Ma-Jian1:
move codegen from outside into Expr, dispatch codegen() by virtual method.

dispatch by dynamic_cast is slow

No

UT

INFRA


cider/type/plan/Analyzer2.cpp
```diff
@@ -0,0 +1,211 @@
...
+ */
+#include "exec/nextgen/jitlib/JITLib.h"
+#include "exec/nextgen/jitlib/base/JITFunction.h"
+#include "exec/template/Execute.h"
+#include "type/plan/Analyzer.h"
+
+namespace Analyzer {
+using namespace cider::jitlib;
+
+JITTypeTag Expr::getJITTag(const SQLTypes& st) {
```
bigPYJ1151, line 29: Maybe we could separate this func to ```nextgen/utils```? So other components could use it.

cider/type/plan/Analyzer.h
```diff
@@ -323,6 +334,9 @@ class ColumnVar : public Expr {
   bool operator==(const Expr& rhs) const override;
   std::string toString() const override;
 
+ public:
+  JITExprValue& codegen(JITFunction& func) override;
```
bigPYJ1151, line 338: Should we move these ```Expr``` subclasses to separate files step-by-step? Maybe ```Analyzer.h``` is oversized.

cider/type/plan/Analyzer2.cpp
Ma-Jian1: only Expr needs this so far. 
We can make this change any time when we found others need this.

cider/type/plan/Analyzer.h
Ma-Jian1: good idea

ci/scripts/intel_copyright_files.txt
```diff
@@ -22,6 +22,10 @@ cider/exec/plan/parser/*Visitor*
 cider/exec/plan/parser/Substrait*
 cider/exec/plan/parser/ConverterHelper.*
 cider/exec/nextgen/*
+cider/type/plan/BinaryExpr.cpp
+cider/type/plan/ColumnExpr.cpp
+cider/type/plan/ConstantExpr.cpp
+cider/type/plan/Expr.cpp
```
bigPYJ1151, line 28: Use regex?

Ma-Jian1: done

bigPYJ1151 (approved):

cider/type/plan/BinaryExpr.cpp
```diff
@@ -0,0 +1,129 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include "BinaryExpr.h"
+#include "exec/template/Execute.h"  // for is_unnest
+
+namespace Analyzer {
+using namespace cider::jitlib;
+
+JITExprValue& BinOper::codegen(JITFunction& func) {
```
xieqi, line 27: These files are related to codegen, and it is not appropriate to put them in the plan directory

winningsix (approved):

cider/type/plan/BinaryExpr.cpp
Ma-Jian1: then should we move all Analyzer out of plan?

winningsix: Hmm, will be better to decouple plan definition and codegen logic? Saying ```ColumnVar``` with ```ColumnVarCodegen```. Thoughts?

Ma-Jian1: we may reconsider how to refactor this API after the nextgen is finished.

PR #188, opened 2022-11-21
[POAE7-2479] Use share_ptr<CiderBatch> for join bridge

yeyuqiang:
Pass UTs with project node after enable join pattern

No

existing UT after enable join pattern manually

veloxIntegration, test

### DIFF
--- a/cider-velox/src/CiderOperator.h
+++ b/cider-velox/src/CiderOperator.h
@@ -61,8 +61,6 @@ class CiderOperator : public exec::Operator {
   std::shared_ptr<DataConvertor> dataConvertor_;
   std::chrono::microseconds convertorInternalCounter;
 
-  // these properties are used for join with w/o agg case
-  std::optional<std::shared_ptr<CiderBatch>> buildData_;
   bool buildSideEmpty_{false};
   bool buildTableFed_{false};
   bool finished_{false};
--- a/cider-velox/src/CiderOperator.h
+++ b/cider-velox/src/CiderOperator.h
@@ -62,11 +62,11 @@ class CiderOperator : public exec::Operator {
   std::chrono::microseconds convertorInternalCounter;
 
   // these properties are used for join with w/o agg case
-  std::optional<CiderBatch> buildData_;
+  std::optional<std::shared_ptr<CiderBatch>> buildData_;
   bool buildSideEmpty_{false};
   bool buildTableFed_{false};
   bool finished_{false};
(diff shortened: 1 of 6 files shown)


cider-velox/test/CiderOperatorHashJoinTest.cpp
```diff
@@ -114,10 +114,11 @@ class CiderOperatorHashJoinTest : public CiderOperatorTestBase {
 TEST_F(CiderOperatorHashJoinTest, innerJoin_allTypes) {
   std::vector<TypePtr> types{BIGINT(), INTEGER()};
   for (const auto& type : types) {
-    testJoin({type},
```
winningsix, line 117: Hmm, a quick question: how to we ensure join is executed by Cider?
```
  void assertPlanConversion(const std::shared_ptr<const core::PlanNode>& plan,
                            const std::string& duckDbSql) {
    auto ciderPlan = CiderVeloxPluginCtx::transformVeloxPlan(plan);
    assertQuery(ciderPlan, duckDbSql);
  }
```
It seems only validating the result which may not be executed on Cider.

yeyuqiang: CiderOp has been registered in presto-cpp logically and controlled by CiderPlanTransformer programatically.

winningsix: In this test, should we pass in some config to control it?

yeyuqiang: We need enable join pattern in CiderPlanTransformer in the future. Current we disable it since there's still some other issues in integration test for join.

winningsix (approved):

cider-velox/src/CiderOperator.h
```diff
@@ -62,11 +62,11 @@ class CiderOperator : public exec::Operator {
   std::chrono::microseconds convertorInternalCounter;
 
   // these properties are used for join with w/o agg case
-  std::optional<CiderBatch> buildData_;
+  std::optional<std::shared_ptr<CiderBatch>> buildData_;
```
chaojun-zhang, line 65: maybe we don't need this member , since we move into cidercompile module?

yeyuqiang: Yes, deleted the buildData_.

chaojun-zhang (approved):

winningsix (approved):

PR #195, opened 2022-11-22
[POAE7-2599] verify PartialAVG with groupby for arrow format

imajiayu:
add Test Case that ParitialAVG with groupby for arrow format

No

CiderAggArrowTest.cpp

### DIFF
--- a/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
+++ b/cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
@@ -78,14 +78,24 @@ class SimpleAggExtractor : public CiderAggTargetColExtractor {
           CiderBitUtils::clearBitAt(nulls, i);
           ++null_count;
         } else {
-          buffer[i] = *reinterpret_cast<const ST*>(rowPtr + offset_);
+          if (this->getName() == "FLOAT_DOUBLE") {
+            buffer[i] =
+                std::stod(std::to_string(*reinterpret_cast<const ST*>(rowPtr + offset_)));
+          } else {
+            buffer[i] = *reinterpret_cast<const ST*>(rowPtr + offset_);
+          }
         }
       }
       output->setNullCount(null_count);
     } else {
       for (size_t i = 0; i < rowNum; ++i) {
         const int8_t* rowPtr = rowAddrs[i];
-        buffer[i] = *reinterpret_cast<const ST*>(rowPtr + offset_);
+        if (this->getName() == "FLOAT_DOUBLE") {
+          buffer[i] =
+              std::stod(std::to_string(*reinterpret_cast<const ST*>(rowPtr + offset_)));
+        } else {
+          buffer[i] = *reinterpret_cast<const ST*>(rowPtr + offset_);
+        }
       }
     }
   }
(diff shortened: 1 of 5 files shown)


cider/tests/utils/CiderQueryRunner.cpp
```diff
@@ -230,11 +230,11 @@ std::vector<CiderBatch> CiderQueryRunner::handleRes(
   std::vector<CiderBatch> res;
   auto has_more_output = CiderRuntimeModule::ReturnCode::kMoreOutput;
   while (has_more_output == CiderRuntimeModule::ReturnCode::kMoreOutput) {
-    std::vector<const int8_t*> out_col_buffers(column_num);
-    for (size_t i = 0; i < column_num; ++i) {
-      size_t type_bytes = schema->GetColumnTypeSize(i);
-      out_col_buffers[i] = new int8_t[type_bytes * max_output_row_num];
-    }
+    // std::vector<const int8_t*> out_col_buffers(column_num);
```
winningsix, line 233: Q: why do we need to comment this out?

yma11: These codes are useless so we should remove it directly. Will update this.

cider/tests/functionality/CiderAggArrowTest.cpp
```diff
@@ -955,6 +955,112 @@ TEST_F(CiderPartialAVGFpArrowTest, mixedPartialAVG) {
...
+  std::vector<int64_t> expect_count4({0, 0, 0});
+
+  ArrowSchema* expect_schema;
+  ArrowArray* expect_array;
+
+  std::tie(expect_schema, expect_array) =
+      ArrowArrayBuilder()
+          .setRowNum(3)
+          .addColumn<double>("", CREATE_SUBSTRAIT_TYPE(Fp64), groupby_key)
+          .addColumn<double>("", CREATE_SUBSTRAIT_TYPE(Fp64), expect_sum1)
```
jikunshang, line 976: should we have a struct arrow_array?

imajiayu: this will be done in the next commit

cider/exec/operator/aggregate/CiderAggTargetColExtractorTmpl.h
```diff
@@ -78,14 +78,24 @@ class SimpleAggExtractor : public CiderAggTargetColExtractor {
           CiderBitUtils::clearBitAt(nulls, i);
           ++null_count;
         } else {
-          buffer[i] = *reinterpret_cast<const ST*>(rowPtr + offset_);
+          if (this->getName() == "FLOAT_DOUBLE") {
+            buffer[i] =
+                std::stod(std::to_string(*reinterpret_cast<const ST*>(rowPtr + offset_)));
```
Deegue, line 83: Q: why we have to transfer ST to string and then to double? same below

imajiayu: from float to double, the direct cast will cause loss of accuracy. For example, 4.6 will be 4.599999904632568, lead to a checker fail

jikunshang: I think it's ok to do this as a work around. But actually we should accept such loss of accuracy if we need do float to double conversion. Otherwise we should use decimal.

yma11: In nextgen, we should won't need to handle these scenarios, which caused by cider internal sum(fp)->fp conflicts with velox sum(fp)->double. We should support support sum(fp)->double directly.

yma11 (approved):

PR #196, opened 2022-11-23
[POAE7-2553] String Scalar Function support: length

YBRua:

Added support for string scalar function `LENGTH(str)`, which computes the number of characters in an input `str`.

Notes:

1. There is an unused class `CharLengthExpr : Expr` in the code base (`Analyzer.h`), but to maintain consistency with existing string operations, this class is *not* used and a new `CharLengthStringOper : StringOper` is used instead.
2. The formal name for this function according to the substrait specification should be `char_length`, but DuckDb only supports `LENGTH`, and therefore all test cases used `LENGTH` for compatibility with duckdb.
3. Isthumus does not support converting `LENGTH()` or `CHAR_LENGTH()`. Current test cases are generated by a modified local fork of Isthmus.

String op support.

No.

UTs.


cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
```diff
@@ -37,8 +37,15 @@ bool getExprUpdatable(std::unordered_map<std::shared_ptr<Analyzer::Expr>, bool>
 }
 
 bool isStringFunction(const std::string& function_name) {
-  std::unordered_set<std::string> supportedStrFunctionSet{
-      "substring", "substr", "lower", "upper", "trim", "ltrim", "rtrim"};
+  std::unordered_set<std::string> supportedStrFunctionSet{"substring",
```
harborn, line 40:
```c++
static std::unordered_set<std::string> funcs = {"substring", ...};
```
1. declare it as a static variable
2. naming the variable as short as it.

cider/function/string/StringOps.cpp
```diff
@@ -100,6 +100,16 @@ Datum TryStringCast::numericEval(const std::string_view str) const {
   }
 }
 
+NullableStrType CharLength::operator()(const std::string& str) const {
+  CIDER_THROW(CiderCompileException, "invalid call to operator() for CharLength");
+}
+
+Datum CharLength::numericEval(const std::string_view str) const {
```
harborn, line 107:
```c++
const std::string_view& str
```

cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
YBRua: updated

cider/function/string/StringOps.cpp
YBRua: updated

jikunshang (approved):

winningsix (approved):

PR #205, opened 2022-11-25
[POAE7-2630] import tpc-ds queries

harborn:


cider-velox/test/e-2-e/presto/queries/tpcds/01.sql
```diff
@@ -0,0 +1,24 @@
+WITH customer_total_return AS
```
winningsix, line 1: Please add licenses for those files.

chaojun-zhang (dismissed):

cider-velox/test/e-2-e/presto/queries/tpcds/01.sql
winningsix: Also, we need to add copyright here like https://github.com/intel/BDTK/blob/main/cider-velox/test/e-2-e/presto/cider-ut/CiderAggTest.sql#LL1C18-L1C39

winningsix (approved):

winningsix (approved):

ZJie1 (approved):

PR #207, opened 2022-11-25
[POAE7-2592] Support function lookup in Substrait2EU

haiweizh:

Support function lookup in Substrait2EU.

Support function lookup in Substrait2EU.

No.

Uts.

CODEREFACTOR,cider,test


cider/function/extensions/substrait/substrait_extension.yaml
```diff
@@ -21,3 +21,127 @@ scalar_functions:
...
+  - 
+    name: "in"
+    description: >-
+      Whether the `expression` is in value list.
+    impls:
+      - args:
+          - value: fp64
+            name: expression
+            description: The expression to test for in the value list.
+          - value: fp64
```
jikunshang, line 118: value should be a list of xx type?

cider/function/extensions/substrait/substrait_extension.yaml
```diff
@@ -21,3 +21,127 @@ scalar_functions:
...
+      - args:
+          - value: fp32
+            name: expression
+            description: The expression to test for in the value list.
+          - value: fp32
+            name: list
+            description: The value list.
+        return: BOOLEAN
+  - 
+    name: "in"
```
jikunshang, line 110: need define string type in function?

cider/function/internals/extension_types.yaml
```diff
@@ -0,0 +1,10 @@
+---
+types:
+  - name: point
```
jikunshang, line 3: necessary?

cider/type/plan/Analyzer.h
```diff
@@ -47,38 +47,42 @@
...
-  kSTRING_OPER,
-  kLOWER_STRING_OPER,
-  kSUBSTRING_STRING_OPER,
-  kFUNCTION_OPER,
-  kOFFSET_IN_FRAGMENT,
-  kWINDOW_FUNCTION,
-  kARRAY_EXPR,
-  kUNDEFINED_EXPR
+  kCOLUMN_VAR = 0,
+  kEXPRESSION_TUPLE = 1,
```
jikunshang, line 51: don't need `=1`, same as below

cider/type/plan/Analyzer.h
```diff
@@ -47,38 +47,42 @@
...
+  kFUNCTION_OPER = 26,
+  kOFFSET_IN_FRAGMENT = 27,
+  kWINDOW_FUNCTION = 28,
+  kARRAY_EXPR = 29,
+  kIS_NOT_NULL = 30,
+  kSUBSTRING_STRING_OPER = 31,
+  kUPPER_STRING_OPER = 32,
+  kTRIM_STRING_OPER = 33,
+  kCONCAT_STRING_OPER = 34,
+  kUNDEFINED_EXPR = 35,
```
jikunshang, line 85: undefined could be -1

cider/exec/plan/parser/ConverterHelper.cpp
```diff
@@ -43,18 +43,18 @@ void registerExtensionFunctions() {
...
+      // get op type, no need mapping in substraitToAnalyzerExpr
+      function_map.emplace(extension.extension_function().function_anchor(), function);
     }
   }
-  return function_map;
+  return std::move(function_map);
 }
 
-std::string getFunctionName(const std::unordered_map<int, std::string> function_map,
+std::string getFunctionName(const std::unordered_map<int, std::string>& function_map,
```
yma11, line 57: So here we get not only the function name? if so, rename it to `getFunction`?

cider/exec/plan/parser/ConverterHelper.cpp
```diff
@@ -283,7 +283,7 @@ substrait::Type getSubstraitType(const SQLTypeInfo& type_info) {
   }
 }
 
-SQLOps getCiderSqlOps(const std::string op) {
+SQLOps getCiderSqlOps(const std::string& op) {
   if (op == "lt") {
```
yma11, line 287: Use a map `<string, SQLOps`?

cider/exec/plan/parser/SubstraitToAnalyzerExpr.cpp
```diff
@@ -1059,9 +1070,28 @@ std::shared_ptr<Analyzer::Expr> Substrait2AnalyzerExprConverter::toAnalyzerExpr(
...
-        expr_map_ptr) {
-  auto function = getFunctionName(function_map, s_expr.function_reference());
-  SQLAgg agg_kind = getCiderAggOp(function);
+        expr_map_ptr,
+    std::string return_type) {
+  auto function_sig = getFunctionName(function_map, s_expr.function_reference());
+  if (return_type.empty()) {
+    return_type = TypeUtils::getStringType(s_expr.output_type());
+  }
+  const auto function_lookup_ptr = FunctionLookupEngine::getInstance(from_platform_);
```
yma11, line 1079: maybe we only maintain one instance in substrait2EU?

jikunshang:
please check why build takes about 4 hours.

cider/function/FunctionLookupEngine.cpp
```diff
@@ -195,15 +283,101 @@ const FunctionDescriptor FunctionLookupEngine::lookupFunction(
...
+}
+
+const FunctionDescriptor FunctionLookupEngine::lookupFunction(
+    const std::string& function_signature_str,
+    const std::string& function_return_type_str,
+    const PlatformType& from_platform) const {
+  FunctionDescriptor function_descriptor;
+  auto function_name =
+      function_signature_str.substr(0, function_signature_str.find_first_of(':'));
+  auto function_args = function_signature_str.substr(
```
yma11, line 327: So here we rely on the correctness of function args types defined in substrait plan. Will it be complicated if we get the exact arg types from the expression?

winningsix:
@ZJie1  Please also help take a look at this PR.

cider/function/extensions/substrait/substrait_extension.yaml
haiweizh: In oper in Substrait json is like: "in:fp64_fp64", so the second arg value is not need to be a list.

cider/function/extensions/substrait/substrait_extension.yaml
haiweizh: String function is in functions_string.yaml.
    name: "in"
    description: >-
      Whether the `expression` is in value list.
    impls:
      - args:
          - value: string
            name: expression
            description: The expression to test for in the value list.
          - value: string
            name: list
            description: The value list.
        return: BOOLEAN

cider/function/internals/extension_types.yaml
haiweizh: Removed.

cider/type/plan/Analyzer.h
haiweizh: This is added for debug. If not the debug log is number, that is inconvenient for comparison.

cider/type/plan/Analyzer.h
haiweizh: Done.

cider/exec/plan/parser/ConverterHelper.cpp
haiweizh: Done. Changed it to getFunctionSignature, which the signature string is used for function lookup.

haiweizh:
> please check why build takes about 4 hours.

This is caused by the fact that the substrait-cpp header file references the substrait pb file which has not been compiled yet. So i have updated the code.

cider/function/FunctionLookupEngine.cpp
haiweizh: Yes, it will be more complicated. So we should timely update the yaml file when update and add the cider function.

haiweizh:
I think we should add a platform argument in CiderRuntimeModule in the future, which is used to identify which platform's function we look up. Now I have written it as a fixed velox argument in the Substrait2EU.

cider/function/FunctionLookupEngine.h
```diff
@@ -34,8 +34,33 @@
 #include "FunctionSignature.h"
 #include "Type.h"
 #include "function/SubstraitFunctionCiderMappings.h"
+#include "include/cider/CiderSupportPlatType.h"
 
-enum PlatformType { SubstraitPlatform, PrestoPlatform, SparkPlatform };
+namespace io::substrait {
+
+class VeloxFunctionMappings : public FunctionMapping {
```
winningsix, line 41: Hmm, it's not velox but either Spark or Presto, right? Velox itself didn't bring definition of function/operator semantics.

haiweizh: Done. Rename to presto.

yma11 (approved):

cider/include/cider/CiderSupportPlatType.h
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef CIDER_CIDERSUPPORTPLATTYPE_H
+#define CIDER_CIDERSUPPORTPLATTYPE_H
+
+enum PlatformType { SubstraitPlatform, PrestoPlatform, SparkPlatform };
```
jikunshang, line 25: rename this file? CiderSupportPlatType.h  -> CiderSupportPlatformType.h

jikunshang (approved):

PR #208, opened 2022-11-25
[POAE7-2627] support STRUCT result for non-groupby agg for arrow format

imajiayu:
support STRUCT result for non-groupby agg for arrow format

No

CiderAggArrowTest

### DIFF
--- a/cider/exec/module/batch/CiderBatchUtils.cpp
+++ b/cider/exec/module/batch/CiderBatchUtils.cpp
@@ -33,40 +33,41 @@
 
 namespace CiderBatchUtils {
 
-#define GENERATE_AND_ADD_BOOL_COLUMN(C_TYPE)                                \
-  {                                                                         \
-    std::vector<C_TYPE> col_data;                                           \
-    std::vector<bool> null_data;                                            \
-    col_data.reserve(table_row_num);                                        \
-    null_data.reserve(table_row_num);                                       \
-    C_TYPE* buf = (C_TYPE*)table_ptr[i];                                    \
-    uint8_t* null_buf = (uint8_t*)table_ptr[i + column_num];                \
-    for (auto j = 0; j < table_row_num; ++j) {                              \
-      C_TYPE value = buf[j];                                                \
-      bool is_not_null = null_buf[j];                                       \
-      col_data.push_back(value);                                            \
-      null_data.push_back(!is_not_null);                                    \
-    }                                                                       \
-    builder = builder.addBoolColumn<C_TYPE>(names[i], col_data, null_data); \
-    break;                                                                  \
+#define GENERATE_AND_ADD_BOOL_COLUMN(C_TYPE)                                            \
+  {                                                                                     \
+    std::vector<C_TYPE> col_data;                                                       \
+    std::vector<bool> null_data;                                                        \
+    col_data.reserve(table_row_num);                                                    \
+    null_data.reserve(table_row_num);                                                   \
+    C_TYPE* buf = (C_TYPE*)table_ptr[table_ptr_idx];                                    \
+    uint8_t* null_buf = (uint8_t*)table_ptr[table_ptr_idx + column_num];                \
+    for (auto j = 0; j < table_row_num; ++j) {                                          \
+      C_TYPE value = buf[j];                                                            \
+      bool is_not_null = null_buf[j];                                                   \
+      col_data.push_back(value);                                                        \
+      null_data.push_back(!is_not_null);                                                \
+    }                                                                                   \
+    builder = builder.addBoolColumn<C_TYPE>(names[table_ptr_idx], col_data, null_data); \
+    break;                                                                              \
   }
 
-#define GENERATE_AND_ADD_COLUMN(C_TYPE)                                       \
-  {                                                                           \
-    std::vector<C_TYPE> col_data;                                             \
-    std::vector<bool> null_data;                                              \
-    col_data.reserve(table_row_num);                                          \
-    null_data.reserve(table_row_num);                                         \
-    C_TYPE* buf = (C_TYPE*)table_ptr[i];                                      \
-    uint8_t* null_buf = (uint8_t*)table_ptr[i + column_num];                  \
-    for (auto j = 0; j < table_row_num; ++j) {                                \
-      C_TYPE value = buf[j];                                                  \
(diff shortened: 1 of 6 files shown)


cider/tests/utils/ArrowArrayBuilder.h
```diff
@@ -43,7 +43,7 @@ class ArrowArrayBuilder {
 
     array_->buffers = (const void**)allocator_->allocate(sizeof(void*));
     array_->buffers[0] = nullptr;
-    array_->n_buffers = 0;
+    array_->n_buffers = 1;
```
winningsix, line 46: Q: why do we need to set to 1 considering ```array_->buffers[0] = nullptr``` in L45?

cider/exec/module/batch/CiderBatchUtils.cpp
```diff
@@ -434,20 +435,14 @@ std::string extractUtf8ArrowArrayAt(const ArrowArray* array, size_t index) {
...
-  auto table_row_num = output_batch.row_num();
-  ArrowArrayBuilder builder;
-  builder = builder.setRowNum(table_row_num);
-  const auto& types = table_schema->getFlattenColumnTypes();
-  const auto& names = table_schema->getFlattenColNames();
-  CHECK_EQ(types.size(), names.size());
-  const int8_t** table_ptr = output_batch.table();
-  for (auto i = 0; i < types.size(); ++i) {
-    const ::substrait::Type& type = types[i];
+int convertToStructArrow(ArrowArrayBuilder& builder,
```
winningsix, line 438: minor hint: ```convertToStructArrow```->  ```convertToArrowStruct```

cider/exec/module/batch/CiderBatchUtils.cpp
```diff
@@ -466,6 +461,36 @@ CiderBatch convertToArrowRepresentation(const CiderBatch& output_batch) {
...
+    table_ptr_num += convertToStructArrow(
+        subBuilder, subType, table_ptr, table_ptr_idx, table_row_num, column_num, names);
+    table_ptr_idx += table_ptr_num;
+  }
+  auto schema_and_array = subBuilder.build();
+  builder.addStructColumn(std::get<0>(schema_and_array), std::get<1>(schema_and_array));
+  return table_ptr_num;
+}
+
+CiderBatch convertToArrowRepresentation(const CiderBatch& output_batch) {
```
winningsix, line 479: Minor hint: ```convertToArrowRepresentation```  simplify to ```convertToArrow```

cider/exec/module/batch/CiderBatchUtils.cpp
```diff
@@ -434,20 +435,14 @@ std::string extractUtf8ArrowArrayAt(const ArrowArray* array, size_t index) {
...
-  for (auto i = 0; i < types.size(); ++i) {
-    const ::substrait::Type& type = types[i];
+int convertToStructArrow(ArrowArrayBuilder& builder,
+                         const ::substrait::Type type,
+                         const int8_t** table_ptr,
+                         int table_ptr_idx,
+                         int table_row_num,
+                         int column_num,
+                         std::vector<std::string> names) {
+  if (!type.has_struct_()) {
```
winningsix, line 445: Q: What if we have struct or other nested type?

cider/tests/utils/ArrowArrayBuilder.h
imajiayu: This buffer is allocated as a bitmap to show which children is NULL or not in **root** node of Arrow format. But it has not been used so far, so we temporarily set it to nullptr, and we should consider the number of buffers is 1. Otherwise, the later check in `StructBatch` will throw an exception.

cider/exec/module/batch/CiderBatchUtils.cpp
imajiayu: This is a recursive function to build nested type. L445 is the terminate condition, which will build a primitive column and return, L467-L476 will build a struct recursively.

cider/exec/module/batch/CiderBatchUtils.cpp
```diff
@@ -466,6 +461,36 @@ CiderBatch convertToArrowRepresentation(const CiderBatch& output_batch) {
...
     }
+    return 1;
+  }
+
+  ArrowArrayBuilder subBuilder;
+  int table_ptr_num = 0;
+  for (auto subType : type.struct_().types()) {
+    table_ptr_num += convertToArrowStruct(
+        subBuilder, subType, table_ptr, table_ptr_idx, table_row_num, column_num, names);
+    table_ptr_idx += table_ptr_num;
```
jikunshang, line 472: is this correct?

cider/tests/utils/ArrowArrayBuilderTest.cpp
```diff
@@ -288,7 +288,7 @@ TEST(ArrowArrayBuilderTest, CiderBatchConstructorTest) {
 
   CiderBatch* batch =
       new CiderBatch(schema, array, std::make_shared<CiderDefaultAllocator>());
-  EXPECT_EQ(0, batch->getBufferNum());
+  EXPECT_EQ(1, batch->getBufferNum());
```
jikunshang, line 291: can you add some ut like build a struct arrowArray/CiderBatch and check whether equal?

cider/exec/module/batch/CiderBatchUtils.cpp
yma11: Should it be same as below L491-L493?

cider/exec/module/batch/CiderBatchUtils.cpp
```diff
@@ -466,6 +461,36 @@ CiderBatch convertToArrowRepresentation(const CiderBatch& output_batch) {
...
+  ArrowArrayBuilder builder;
+  builder = builder.setRowNum(table_row_num);
+  const auto& types = table_schema->getColumnTypes();
+  const auto& names = table_schema->getFlattenColNames();
+  const int8_t** table_ptr = output_batch.table();
+  int table_ptr_idx = 0;
+  for (auto type : types) {
+    table_ptr_idx += convertToArrowStruct(
+        builder, type, table_ptr, table_ptr_idx, table_row_num, column_num, names);
   }
```
yma11, line 494: Can we add some comments about the struct batch convert here? like how it's constructed with schema/table_ptr given.

PR #210, opened 2022-11-28
[POAE7-2648] support null in filter and expr

Ma-Jian1:
support null in filter and expr
disable implicit cast for bool for some operator overload.

next-gen compiler

No

UT

INFRA

### DIFF
--- a/cider/type/plan/BinaryExpr.cpp
+++ b/cider/type/plan/BinaryExpr.cpp
@@ -49,12 +49,12 @@ JITExprValue& BinOper::codegen(JITFunction& func) {
   FixSizeJITExprValue rhs_val(rhs->codegen(func));
 
   JITValuePointer null(nullptr);
-  auto lhs_null = lhs_val.getNull().get();
-  auto rhs_null = rhs_val.getNull().get();
-  if (lhs_null && rhs_null) {
-    null = *lhs_null || *rhs_null;
-  } else if (lhs_null || rhs_null) {
-    null = lhs_null ? lhs_null : rhs_null;
+  bool lhs_nullable = lhs_val.isNullable();
+  bool rhs_nullable = rhs_val.isNullable();
+  if (lhs_nullable && rhs_nullable) {
+    null.replace(lhs_val.getNull() || rhs_val.getNull());
+  } else if (lhs_nullable || rhs_nullable) {
+    null.replace(lhs_nullable ? lhs_val.getNull() : rhs_val.getNull());
   }
 
   switch (lhs_ti.get_type()) {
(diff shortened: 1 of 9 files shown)


cider/type/plan/ConstantExpr.cpp
```diff
@@ -34,7 +34,8 @@ JITExprValue& Constant::codegen(JITFunction& func) {
       CIDER_THROW(CiderCompileException,
                   "NULL type literals are not currently supported in this context.");
     case kBOOLEAN:
-      return set_expr_value(func.createConstant(getJITTag(type), get_constval().boolval));
+      return set_expr_value(nullptr,
```
yma11, line 37: `nullptr` should be `func.createConstant(getJITTag(type), false)` to indicate it's a not null constant.

cider/type/plan/BinaryExpr.cpp
```diff
@@ -86,41 +79,45 @@ JITExprValue& BinOper::codegen(JITFunction& func) {
...
 
-JITExprValue& BinOper::codegenFixedSizeColCmpFun(JITValue& lhs, JITValue& rhs) {
+JITExprValue& BinOper::codegenFixedSizeColCmpFun(JITValuePointer& null,
+                                                 JITValue& lhs,
+                                                 JITValue& rhs) {
   // TODO: Null Process
   switch (get_optype()) {
     case kEQ:
-      return set_expr_value(nullptr, lhs == rhs);
+      return set_expr_value(null, lhs == rhs);
```
yma11, line 110: maybe we should refine this by adding a new API `set_expr_null` to avoid passing the null everywhere?

yma11: maybe you should use `FixSizeJITExprValue`.

cider/type/plan/ConstantExpr.cpp
Ma-Jian1: It's ok to create a false constant, but then we will need to check null in IR. any suggestions? @bigPYJ1151

cider/type/plan/BinaryExpr.cpp
Ma-Jian1: It's just null value, should be passed whether it is nullptr or not.

cider/type/plan/ConstantExpr.cpp
bigPYJ1151: I think the nullability of Constant should always be false except Constant with kNULLT type. So we could pass a nullptr in most cases.

yma11: If we set the null to `nullptr`, it probably cause problem when trying to use `expr.getNull()` when we don't know its a constant or not.

Ma-Jian1: I add an API `isNullable()`, we should always check it

yma11 (approved):

PR #211, opened 2022-11-29
[POAE7-2561] String Scalar Function support: regexp_replace

YBRua:

Added support for regular expression function `REGEXP_REPLACE(input, pattern, replacement, position, occurrence)`, which replaces one or more substrings that match `pattern` in `input` with `replacement`.

The implementation is based on substrait specification on `regexp_replace`, which should generally be the same as the `regexp_replace` implemented in PrestoDb (but with additional features to specify `position` and `occurrence`)

- Enabled existing `regexp_replace` code in code base.
- Added unit tests that cover basic features specified by substrait.
- There are 3 optional arguments that controls case sensitivity, multi-line replacement and dot character (`'.'`) matching respectively. These 3 arguments are NOT supported for now for lack of support to parse optional arguments.

String op feature support.

No.

UTs. All query plans are currently manually crafted, because Isthmus does not support `regexp_replace`. Some of the test cases use hand-crafted reference outputs because some features are not supported in DuckDb.

### DIFF
--- a/cider/type/plan/Analyzer.cpp
+++ b/cider/type/plan/Analyzer.cpp
@@ -431,6 +431,25 @@ std::shared_ptr<Analyzer::Expr> ConcatStringOper::deep_copy() const {
       std::dynamic_pointer_cast<Analyzer::StringOper>(StringOper::deep_copy()));
 }
 
+std::shared_ptr<Analyzer::Expr> RegexpReplaceStringOper::deep_copy() const {
+  return makeExpr<Analyzer::RegexpReplaceStringOper>(
+      std::dynamic_pointer_cast<Analyzer::StringOper>(StringOper::deep_copy()));
+}
+
+std::vector<std::shared_ptr<Analyzer::Expr>> RegexpReplaceStringOper::foldLiteralStrCasts(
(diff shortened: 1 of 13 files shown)


cider/type/plan/Analyzer.cpp
```diff
@@ -431,6 +431,32 @@ std::shared_ptr<Analyzer::Expr> ConcatStringOper::deep_copy() const {
...
 
+std::shared_ptr<Analyzer::Expr> RegexpReplaceStringOper::deep_copy() const {
+  return makeExpr<Analyzer::RegexpReplaceStringOper>(
+      std::dynamic_pointer_cast<Analyzer::StringOper>(StringOper::deep_copy()));
+}
+
+std::vector<std::shared_ptr<Analyzer::Expr>> StringOper::foldLiteralStrCasts(
+    const std::vector<std::shared_ptr<Analyzer::Expr>>& operands,
+    int start_idx) {
+  std::vector<std::shared_ptr<Analyzer::Expr>> folded_operands;
```
harborn, line 442: should `reserve` a capacity

jikunshang (approved):

winningsix (approved):

PR #213, opened 2022-11-29
[POAE7-2624] support is distinct from for nextgen

yma11:
support IS DISTINCT FROM codegen

support IS DISTINCT FROM codegen

No

UT

### DIFF
--- a/cider/type/plan/BinaryExpr.cpp
+++ b/cider/type/plan/BinaryExpr.cpp
@@ -20,10 +20,59 @@
  */
 #include "type/plan/BinaryExpr.h"
 #include "exec/nextgen/jitlib/base/JITValue.h"
+#include "exec/plan/parser/ParserNode.h"
 #include "exec/template/Execute.h"  // for is_unnest
 
 namespace Analyzer {
 using namespace cider::jitlib;
+
+std::shared_ptr<Analyzer::BinOper> lower_bw_eq(const Analyzer::BinOper* bw_eq) {
+  const auto eq_oper =
+      std::make_shared<Analyzer::BinOper>(bw_eq->get_type_info(),
+                                          bw_eq->get_contains_agg(),
+                                          kEQ,
+                                          bw_eq->get_qualifier(),
+                                          bw_eq->get_own_left_operand(),
+                                          bw_eq->get_own_right_operand());
+  const auto lhs_is_null = std::make_shared<Analyzer::UOper>(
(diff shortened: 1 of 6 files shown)


cider/type/plan/BinaryExpr.cpp
```diff
@@ -121,8 +122,56 @@ JITExprValue& BinOper::codegenFixedSizeColCmpFun(JITValuePointer& null,
...
+                 lhs_val.getValue() != rhs_val.getValue()) ||
+                (lhs_val.getNull() != rhs_val.getNull());
+      } else if (lhs_val.isNullable()) {
+        value = lhs_val.getValue() != rhs_val.getValue() || lhs_val.getNull();
+      } else if (rhs_val.isNullable()) {
+        value = lhs_val.getValue() != rhs_val.getValue() || rhs_val.getNull();
+      }
+      return set_expr_value(func.createConstant(JITTypeTag::BOOL, false), value);
+    }
+    case kBW_EQ: {
```
bigPYJ1151, line 158: Seems this implementation could be simplified. For example, result of ```kBW_EQ``` could be calculated from the ```NOT``` of the result of ```kBW_NE```.

yma11: The original implementation of `lower_bw_eq` and `lower_bw_ne` has separate logic. But in semantic, we can leverage the `NOT` op. Updated it and will remove the nullable check after [PR](https://github.com/intel/BDTK/pull/218/) merged which should be more straight forward.

cider/type/plan/BinaryExpr.cpp
```diff
@@ -121,8 +122,56 @@ JITExprValue& BinOper::codegenFixedSizeColCmpFun(JITValuePointer& null,
...
   }
+}
 
-  return expr_var_;
+JITExprValue& BinOper::codegenFixedSizeLogicalFun(JITValuePointer& null,
+                                                  JITValue& lhs,
+                                                  JITValue& rhs) {
+  switch (get_optype()) {
+    case kAND:
+      return set_expr_value(null, lhs && rhs);
```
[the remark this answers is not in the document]
yma11: Yes. You are correct. This method is actually not triggered and I will submit a new PR for it with UT covered.

yma11:
@bigPYJ1151 @jikunshang @kd97 please help review this PR. Thanks.

bigPYJ1151 (approved):

jikunshang (approved):

PR #214, opened 2022-11-30
[POAE7-2625] support primitive type cast func in nextgen

kd97:

 support primitive type cast func in nextgen

nextgen compiler

No

new UTs

### DIFF
--- a/cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
+++ b/cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
@@ -317,37 +317,27 @@ JITValuePointer LLVMJITValue::castPointerSubType(JITTypeTag type_tag) {
       JITTypeTag::POINTER, parent_function_, new_llvm_value, "cast_ptr", false, type_tag);
 }
 
-JITValuePointer LLVMJITValue::castJITValuePrimitiveType(JITTypeTag target_jit_tag,
-                                                        bool is_signed) {
+JITValuePointer LLVMJITValue::castJITValuePrimitiveType(JITTypeTag target_jit_tag) {
   llvm::Type* source_type =
       getLLVMType(getValueTypeTag(), parent_function_.getLLVMContext());
   llvm::Type* target_type =
       getLLVMType(target_jit_tag, parent_function_.getLLVMContext());
+  CHECK(source_type && target_type);
   llvm::Value* source_lv = load();
   llvm::Value* target_lv = nullptr;
   if (source_type->isIntegerTy() && target_type->isIntegerTy()) {
-    target_lv = getFunctionBuilder(parent_function_)
-                    .CreateIntCast(source_lv, target_type, is_signed);
+    target_lv =
+        getFunctionBuilder(parent_function_).CreateIntCast(source_lv, target_type, true);
   } else if (source_type->isIntegerTy() && target_type->isFloatingPointTy()) {
-    if (is_signed) {
-      target_lv =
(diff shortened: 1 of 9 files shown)


cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
```diff
@@ -317,6 +317,45 @@ JITValuePointer LLVMJITValue::castPointerSubType(JITTypeTag type_tag) {
       JITTypeTag::POINTER, parent_function_, new_llvm_value, "cast_ptr", false, type_tag);
 }
 
+JITValuePointer LLVMJITValue::castJITValuePrimitiveType(JITTypeTag target_jit_tag,
+                                                        bool is_signed) {
```
bigPYJ1151, line 321: Seems ```is_signed``` could be get from ```source_type```. It's unnecessary to pass it as a parameter.

cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
```diff
@@ -317,6 +317,45 @@ JITValuePointer LLVMJITValue::castPointerSubType(JITTypeTag type_tag) {
...
+  llvm::Type* source_type =
+      getLLVMType(getValueTypeTag(), parent_function_.getLLVMContext());
```
bigPYJ1151, line 323: Return value from ```getLLVMType``` should be checked because it could be ```nullptr```.

cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
kd97: 1. so need to keep this parameter in the `JITValue` method for other compiler impl?  now already set it as a default parameter in `LLVMJITValue.h` file.
2. ` getLLVMType` method will get `llvm::Type` from `JITTypeTag`, seems the specified default type is signed, so can just remove unsigned-cast part?

bigPYJ1151: - I think needn't. Because we can get all type attributes (signed, float, width...) from ```JITTypeTag``` itself, by some helper functions.
 - Currently unsigned integer aren't considered as primitive type in jitlib to keep alignment with original omnisci's type system. We could support unsigned integer easily if needed.

cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
```diff
@@ -317,37 +317,26 @@ JITValuePointer LLVMJITValue::castPointerSubType(JITTypeTag type_tag) {
       JITTypeTag::POINTER, parent_function_, new_llvm_value, "cast_ptr", false, type_tag);
 }
 
-JITValuePointer LLVMJITValue::castJITValuePrimitiveType(JITTypeTag target_jit_tag,
-                                                        bool is_signed) {
+JITValuePointer LLVMJITValue::castJITValuePrimitiveType(JITTypeTag target_jit_tag) {
   llvm::Type* source_type =
       getLLVMType(getValueTypeTag(), parent_function_.getLLVMContext());
   llvm::Type* target_type =
       getLLVMType(target_jit_tag, parent_function_.getLLVMContext());
```
bigPYJ1151, line 324: ```source_type``` and ```target_type``` should be checked here, or ```source_type->isIntegerTy() && target_type->isIntegerTy()``` will core dump when they are nullptr.

bigPYJ1151 (approved):

yma11 (approved):

PR #216, opened 2022-11-30
[POAE7-2647] support STRUCT result for groupby agg for arrow format

imajiayu:
support STRUCT result for groupby agg for arrow format.
Specifically handle groupby Partial AVG in fetch result stage.

No

CiderAggArrowTest.cpp

### DIFF
--- a/cider/exec/module/CiderRuntimeModule.cpp
+++ b/cider/exec/module/CiderRuntimeModule.cpp
@@ -34,6 +34,7 @@
 #include "exec/plan/parser/TypeUtils.h"
 #include "exec/template/CountDistinct.h"
 #include "exec/template/OutputBufferInitialization.h"
+#include "tests/utils/ArrowArrayBuilder.h"
 
 using agg_query = void (*)(const int8_t***,  // col_buffers
                            const uint64_t*,  // num_fragments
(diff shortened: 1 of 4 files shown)


cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -888,6 +889,40 @@ CiderBatch CiderRuntimeModule::setSchemaAndUpdateAggResIfNeed(
         }
       }
     }
+  } else if (ciderCompilationOption_.use_cider_data_format) {
```
jikunshang, line 892: if (ciderCompilationOption_.use_cider_data_format && is_group_by_) would be better?

cider/include/cider/batch/CiderBatch.h
```diff
@@ -115,6 +115,14 @@ class CiderBatch {
 
   std::string toStringForArrow() const;
 
+  ArrowSchema* getArrowSchema() const { return arrow_schema_; }
+
+  ArrowArray* getArrowArray() const { return arrow_array_; }
+
+  ArrowSchema* setArrowSchema(ArrowSchema* schema) { arrow_schema_ = schema; }
```
jikunshang, line 122: I am afraid this method may be unsafe.

imajiayu: I will use CiderBatch Constructor

cider/include/cider/batch/CiderBatch.h
```diff
@@ -115,6 +115,12 @@ class CiderBatch {
 
   std::string toStringForArrow() const;
 
+  ArrowSchema* getArrowSchema() const { return arrow_schema_; }
+
+  ArrowArray* getArrowArray() const { return arrow_array_; }
+
+  void removeOwnerShip() { this->ownership_ = false; }
```
jikunshang, line 122: add some comments here?

jikunshang (dismissed):

cider/include/cider/batch/CiderBatch.h
```diff
@@ -115,6 +115,14 @@ class CiderBatch {
 
   std::string toStringForArrow() const;
 
+  ArrowSchema* getArrowSchema() const { return arrow_schema_; }
+
+  ArrowArray* getArrowArray() const { return arrow_array_; }
```
harborn, line 120: modify same

cider/tests/functionality/CiderAggArrowTest.cpp
```diff
@@ -1019,27 +1019,53 @@ TEST_F(CiderPartialAVGFpArrowTest, mixedGroupbyPartialAVG) {
   std::vector<int64_t> expect_count3({0, 1, 1});
   std::vector<int64_t> expect_count4({0, 0, 0});
 
-  ArrowSchema* expect_schema;
-  ArrowArray* expect_array;
+  ArrowSchema** expect_schemas = (ArrowSchema**)malloc(4 * sizeof(ArrowSchema*));
+  ArrowArray** expect_arrays = (ArrowArray**)malloc(4 * sizeof(ArrowArray*));
```
harborn, line 1023: maybe this two variable cause memory leak?
emm, this is tests codes, maybe not consider too much...

cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -34,6 +34,7 @@
 #include "exec/plan/parser/TypeUtils.h"
 #include "exec/template/CountDistinct.h"
 #include "exec/template/OutputBufferInitialization.h"
+#include "tests/utils/ArrowArrayBuilder.h"
```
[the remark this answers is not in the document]
imajiayu: This builder is also used in 'CiderBatchUtils.cpp' for non-groupby agg, maybe we should move it out from tests/utils.

cider/include/cider/batch/CiderBatch.h
```diff
@@ -115,6 +115,14 @@ class CiderBatch {
 
   std::string toStringForArrow() const;
 
+  ArrowSchema* getArrowSchema() const { return arrow_schema_; }
```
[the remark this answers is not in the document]
imajiayu: there are a lot of codes need to modify if we the change the return type from ptr to class

cider/exec/module/CiderRuntimeModule.cpp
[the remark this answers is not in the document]
yma11: Let's check out whether we still need such logic in nextgen. If not, for arrow support, let's keep it like this.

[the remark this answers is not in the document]
yma11: Confirmed with @bigPYJ1151 offline that we will can generate struct array during codegen directly and current logic won't be needed. So let's keep this for the time being.

yma11 (approved):

PR #219, opened 2022-12-01
[POAE7-2562] String scalar function support: split

YBRua:

Added partial support for the family of SQL `split()` functions.

- Added (partial) support for substrait `string_split(input, separator)` function.
  - Note that current implementation only supports special cases where `string_split()` is used *in conjunction with an immediate array element access* (e.g. `split(col, ',')[2]`) because `string_split()` returns a list of strings by definition, but lists are not support yet.
  - This is done by adding a special case handling in the selection expr when creating EU (because list access is a `FieldReference` (selection) in substrait).
  - Based on an observation that `split()` usually (but not always) comes with an index.
- Added (partial) support for prestodb `split(input, delimiter, limit)` function, which returns a list containing at most `limit` strings, and the last string always contains all remaining characters. Its signature is defined as a presto extension for future use in function lookup.
  - Due to the same reason above, `split()` currently must be used together with a list element access.
- Added support for prestodb `split_part(input, delimiter, part)` function. Also defined as a presto extension.
- Updated runtime functions to support `limit` parameter in prestodb dialect.
- Fixed several bugs in split part runtime function according to unit tests.

References: [PrestoDb Documentation](https://prestodb.io/docs/current/functions/string.html).

String op feature support

No.

UTs. Current tests use hand-crafted substrait query plans.


jikunshang (approved):
LGTM!

yma11 (approved):

PR #220, opened 2022-12-01
[POAE7-2657] batchProcessor interface to support stateless/stateful pipeline without join

chaojun-zhang:

- refactor CiderPipelineOperator as a more cleaner wrapper of BatchProcessor
- refactor BatchProcessor as  pure virtual interface and move member variable into new template class DefaultBatchProcessor, so no need to expose more header classes.
- refactor StatelessProcessor and StatefulProcessor to inherit from DefaultBatchProcessor

 batchProcessor to support stateless pipeline without join

no

wait for test when cider runtime module support the next gen compiler.

cider, veloxIntegration, publicApi


chaojun-zhang:
@xieqi @winningsix  please have to review

cider/exec/plan/substrait/SubstraitPlan.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "SubstraitPlan.h"
+
+namespace cider::plan {
+
+SubstraitPlan::SubstraitPlan(const substrait::Plan& plan) : plan_(plan) {}
```
winningsix, line 26: Minor hint: IMO, better to call it ```SubstraitPlanUtils``` and make it static accepting substrait plan as parameter. Alternatively, caching those results as ```containsAgg``` or ```containsJoin``` locally avoiding further visit.

cider/exec/plan/substrait/SubstraitPlan.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+bool SubstraitPlan::hasAggregateRel() const {
```
winningsix, line 28: Better name ```hasStatefulOp```? More important, how to differentiate group-by and non-group-by? For non-group-by, we need to return true even it's typed as ```has_aggregate```.

cider/exec/plan/substrait/SubstraitPlan.cpp
chaojun-zhang: caching result is make sense

cider/exec/plan/substrait/SubstraitPlan.cpp
chaojun-zhang: stateful is only the concept in batchProcessor, for this utitily class it would be better to follow the same name  inside substrait plan

cider/exec/plan/substrait/SubstraitPlan.cpp
chaojun-zhang: caching result is make sense, but our substrait plan only have one root , so it seems we can't get any benefit  by caching the result

winningsix (approved):

cider-velox/src/CiderPipelineOperator.cpp
```diff
@@ -27,29 +27,22 @@
...
-    batchProcessor_->processNextBatch(std::make_shared<CiderBatch>(inBatch));
+  for (size_t i = 0; i < input->childrenSize(); i++) {
+    input->childAt(i)->mutableRawNulls();
   }
+  ArrowArray* inputArrowArray = CiderBatchUtils::allocateArrowArray();
+  exportToArrow(input, *inputArrowArray);
+  ArrowSchema* inputArrowSchema = CiderBatchUtils::allocateArrowSchema();
+  exportToArrow(input, *inputArrowSchema);
+
+  auto allocator = std::make_shared<PoolAllocator>(operatorCtx_->pool());
```
xieqi, line 42: move to construct?

PR #221, opened 2022-12-05
[POAE7-2644] Add Nextgen Runtime Module

xieqi:

Add nextgen runtime support

Add nextgen runtime support for test

No

Use exist UTs

cider


cider/exec/module/CiderRuntimeModule.cpp
```diff
@@ -179,40 +179,43 @@ CiderRuntimeModule::CiderRuntimeModule(
...
-  is_group_by_ = rel_alg_exe->groupby_exprs.size() && rel_alg_exe->groupby_exprs.front();
-  if (is_group_by_) {
-    initCiderAggGroupByHashTable();
-    initCiderAggTargetColExtractors();
-    group_by_agg_iterator_ = nullptr;
-  }
-
-  std::vector<TargetInfo> target_infos = target_exprs_to_infos(
-      rel_alg_exe->target_exprs, *ciderCompilationResult_->impl_->query_mem_desc_);
+  if (!ciderCompilationOption_.use_nextgen_compiler) {
```
winningsix, line 182: Hmm, for next gen, why not creating different instances separately? Like, we introduce a factory and instantiate impl. (either template-based or next-gen) via compilation option. It will help: (1) easily maintain two impl. moving forward; (2) easier to deprecate old template-based one, thoughts?

cider/exec/nextgen/Nextgen.cpp
```diff
@@ -20,35 +20,43 @@
...
+#include "jitlib/base/JITFunction.h"
 
 namespace cider::exec::nextgen {
 
-std::unique_ptr<context::CodegenContext> compile(const RelAlgExecutionUnit& ra_exe_unit,
-                                                 const jitlib::CompilationOptions& co) {
-  auto context = std::make_unique<context::CodegenContext>();
-  jitlib::LLVMJITModule module("codegen", true, co);
+std::pair<std::unique_ptr<context::RuntimeContext>,
+          std::unique_ptr<context::CodegenContext>>
```
yma11, line 29: It seems weird to return a `CodegenContext` after compile done. What will be included in it? Can the included info be moved to `RuntimeContext`?

cider/tests/nextgen/compiler/NextgenCompilerTest.cpp
```diff
@@ -131,7 +129,7 @@ class CiderNextgenCompilerTestBase : public CiderTestBase {
 };
 
 TEST_F(CiderNextgenCompilerTestBase, integerFilterTest) {
-  assertQueryArrow("SELECT col_1 + col_2 FROM test WHERE col_1 < col_2");
+  assertQueryArrow("SELECT col_1 + col_2 FROM test WHERE col_1 <= col_2");
```
yma11, line 132: Why need this change?

xieqi: due to data pattern, just to make sure the query has result

cider/exec/nextgen/Nextgen.cpp
xieqi: In my test, it will core dump if get RuntimeContext from CodegenContext after call compile function

Ma-Jian1: cause you move `LLVMModule` into `CodegenContext`, maybe should move into `LLVMJITFunction`

cider/exec/nextgen/Nextgen.cpp
```diff
@@ -20,35 +20,43 @@
  */
 
 #include "exec/nextgen/Nextgen.h"
+#include <memory>
```
bigPYJ1151, line 23: Be aware about code style pls.

cider/exec/nextgen/Nextgen.h
```diff
@@ -24,11 +24,17 @@
 #include "exec/nextgen/context/RuntimeContext.h"
 #include "exec/nextgen/parsers/Parser.h"
 #include "exec/nextgen/transformer/Transformer.h"
+#include "jitlib/llvmjit/LLVMJITModule.h"
```
bigPYJ1151, line 27: remove it

cider/exec/nextgen/context/CodegenContext.h
```diff
@@ -22,6 +22,7 @@
 #define NEXTGEN_CONTEXT_CODEGENCONTEXT_H
 
 #include "exec/nextgen/jitlib/base/JITValue.h"
+#include "exec/nextgen/jitlib/llvmjit/LLVMJITModule.h"
```
bigPYJ1151, line 25: remove it

cider/exec/nextgen/context/CodegenContext.h
```diff
@@ -71,21 +71,24 @@ class CodegenContext {
         : ctx_id(id), name(n), type(t) {}
   };
 
-  using BatchDescriptorPtr = std::shared_ptr<BatchDescriptor>;
+  void setJITModule(std::unique_ptr<jitlib::LLVMJITModule> jit_module) {
```
bigPYJ1151, line 74: We should always use base class jitlib::JITModule. BTW, if JITModule will be passed between multiple functions, better to use shared_ptr, same as ```JITFunctionPtr```.

cider/exec/nextgen/jitlib/base/JITModule.h
```diff
@@ -86,7 +86,7 @@ class JITFunctionBuilder {
 
     JITFunctionPointer function = module_->createJITFunction(JITFunctionDescriptor{
         .function_name = name_, .ret_type = return_, .params_type = parameter_list_});
-    builder_(function.get());
+    builder_(function);
```
bigPYJ1151, line 89: Why changed ```builder_```'s signature to shared_ptr? In ```JITFunctionBuilder```, ```builder_``` will be called here only, and always at the same block with ```function```, so there is no lifetime problem.

cider/exec/nextgen/jitlib/base/JITModule.h
```diff
@@ -97,7 +97,7 @@ class JITFunctionBuilder {
   static constexpr size_t DefaultParamsNum = 8;
   boost::container::small_vector<JITFunctionParam, DefaultParamsNum> parameter_list_;
   JITFunctionParam return_;
-  std::function<void(JITFunction*)> builder_;
+  std::function<void(JITFunctionPointer)> builder_;
```
bigPYJ1151, line 100: Same as above.

cider/exec/nextgen/jitlib/base/JITModule.h
xieqi: As builder only save the raw pointer of function, so the function will be released after the build, need to save the function somewhere

cider/exec/nextgen/context/CodegenContext.h
xieqi: module is internal used only

cider/exec/nextgen/jitlib/base/JITModule.h
bigPYJ1151: I see. Maybe change function signature to ```void(const jitlib::JITFunctionPointer&)``` is better.

jikunshang:
per discussed, shouldn't we create BatchProcessor (which leverage nextgen) to replace Compile/Runtime module?

xieqi:
> per discussed, shouldn't we create BatchProcessor (which leverage nextgen) to replace Compile/Runtime module?

yes, in another PR, this for expr codegen

bigPYJ1151 (approved):

yma11 (approved):

PR #224, opened 2022-12-05
[POAE7-2670] [Hot-fix]modify JITValue Logical operations return type

qiuyangshen:
modify JITValue Logical operations return type

bug-fix

No

UTs

FIX


winningsix:
Can we add some tests to capture this bug?

qiuyangshen:
> Can we add some tests to capture this bug?

OK, because It will be implicitly converted to bool , I have added the check in test

Ma-Jian1 (approved):

bigPYJ1151 (approved):

PR #235, opened 2022-12-07
[POAE7-2677] HashTable implemented based on the design of ClickHouse

Deegue:

Migration from ClickHouse.

List of main changes:
1.`FixedHashMap` is ready, which is designed for int8 and int16.
2.Added `CiderNewHashTableTest.cpp` and validated the usage of `FixedHashMap`.
3.Removed and rewrote some dialects under C++20(like `concept`, `requires` etc.)
4.Multi-tread model and buffer reader/writer is disabled currently since we are under the thread model of velox.
5.Memory Arena specially designed for strings is leaving dummy since it's not suitable for cider and we have to implement it later.
6.Memory allocator is now using `CiderDefaultAllocator`, which should be replaced to an allocator with better performance and control later.

Implement new HashTables.

No.

UTs.


cider/common/Arena.h
```diff
@@ -0,0 +1,96 @@
+#pragma once
+
```
harborn, line 2: 直接拷贝过来么？可以thirdparty么？
如果以一定要拷贝过来，这些代码，我理解不应该放在cider下面

Deegue: Several changes have been made as I listed above, and it's not possible to use as thirdparty directly.

spevenhe: Should we move these classes one by one with tests ?

jikunshang (commented):
What's your public interface to other module? `FixedImplicitZeroHashMapWithCalculatedSize` `FixedImplicitZeroHashMapWithCalculatedSize`?

spevenhe:
For license check failed, you may need to add a new header file in ci/scripts and change the license-header.py

cider/common/Arena.h
Deegue: Yes, it should be, but it will result in a lot of PRs.
Here I migrated them at once and we only need to make some adjustment and add test cases later.

Deegue:
> What's your public interface to other module? `FixedImplicitZeroHashMapWithCalculatedSize` `FixedImplicitZeroHashMapWithCalculatedSize`?

Public api will be divided into two parts - agg and join, which are separated from hash table implementation and I'm working on it.

cider/common/Arena.h
harborn: if ClickHouse has changed/updated/iterated this codes, and API or performance optimized, how could we insure those changes effect to cider?
maybe we could a create a new repo, only extract ClickHouse‘s basic code we used.

Deegue: Our aim is to make a high performance hashtable as a universal lib, and it's our first step to migrate ck. For now, we have to sync occasionally, and cherry pick commits good for performance. Later if all functions and public APIs are ready, users will use cider instead of any repo else.

Deegue:
> > > What's your public interface to other module? `FixedImplicitZeroHashMapWithCalculatedSize` `FixedImplicitZeroHashMapWithCalculatedSize`?
> > 
> > 
> > Public api will be divided into two parts - agg and join, which are separated from hash table implementation and I'm working on it.
> 
> I think this part is first thing we need:)

Yes, both are important, and public api for agg will be ready no later than Monday of next week.

cider/common/Arena.h
harborn: ok

cider/common/Types.h
```diff
@@ -0,0 +1,85 @@
...
+#include <string>
+#include <vector>
+
+namespace DB {
+
+/// Data types for representing elementary values from a database in RAM.
+
+/// Hold a null value for untyped calculation. It can also store infinities to handle
+/// nullable comparison which is used for nullable KeyCondition.
+struct Null {
```
Ma-Jian1, line 17: where does this been used?

cider/common/base/types.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
yma11, line 1: Seems it has type system, do we need map cider internal types to these types?

cider/common/base/wide_integer.h
```diff
@@ -0,0 +1,327 @@
...
+    const integer<Bits, Signed>& lhs,
+    const integer<Bits2, Signed2>& rhs);
+template <typename Integral,
+          typename Integral2,
+          class = _only_integer<Integral, Integral2>>
+std::common_type_t<Integral, Integral2> constexpr operator&(const Integral& rhs,
+                                                            const Integral2& lhs);
+
+template <size_t Bits, typename Signed, size_t Bits2, typename Signed2>
+std::common_type_t<integer<Bits, Signed>, integer<Bits2, Signed2>> constexpr operator|(
```
yma11, line 223: I see there are a lot of operator rewrite against the data types, will there be any chance for us to use it?

jikunshang:
should not name this PR as `Migration`

Deegue:
Offline discussed with @jikunshang @yma11 , I will divide this PR into several PRs for different types to make it clear and readable.

Deegue:
Thanks for all reviews above~

PR #240, opened 2022-12-08
[POAE7-2696] Fix memory leakage in FunctionLookupEngine

haiweizh:

Fix memory leakage in FunctionLookupEngine. These memory will be recoveryed until exit by os, that is not grace. Use smart pointer to manage these memory. Cider is a single thread and usually does not cause multiple new, add lock to prevent memory leakage when cider's thread model change.

Fix memory leakage in FunctionLookupEngine

No

Uts

BUG


cider/function/FunctionLookupEngine.cpp
```diff
@@ -24,15 +24,24 @@
...
 
-const FunctionLookupEngine* FunctionLookupEngine::getInstance(
+std::mutex FunctionLookupEngine::s_mutex_;
+
+FunctionLookupEnginePtr FunctionLookupEngine::getInstance(
     const PlatformType from_platform) {
   if (from_platform == PlatformType::SubstraitPlatform ||
       from_platform == PlatformType::PrestoPlatform) {
-    if (function_lookup_engine_ptr_map_.find(from_platform) ==
+    if (function_lookup_engine_ptr_map_.find(from_platform) !=
```
jikunshang, line 33: can remove L33-36 since you add lock on L37?

cider/function/FunctionLookupEngine.h
```diff
@@ -152,8 +155,7 @@ class FunctionLookupEngine {
 
   const PlatformType from_platform_;
   static FunctionLookupEnginePtrMap function_lookup_engine_ptr_map_;
+  static std::mutex s_mutex_;
```
jikunshang, line 158: what's s mean? make it more meaningful?

cider/function/FunctionLookupEngine.cpp
haiweizh: This is to avoid locking to improve performance when the single instance has been initialized.

yma11 (approved):

jikunshang (approved):

PR #245, opened 2022-12-12
[POAE7-2699] [HotFix] Fix input column index in ArrowSourceOp.

bigPYJ1151:

Fix input column index in ArrowSourceOp.

ArrowSourceOp generated wrong index when loading child input ArrowArray.

No

UTs

BugFix

Pls take a review @kd97 @jikunshang


jikunshang (approved):
LGTM. would be better if you can provide a test case.

kd97 (approved):

PR #247, opened 2022-12-12
[POAE7-2677] HashTable implemented with int8 and int16

Deegue:
Support HashTable for key type of int8 and int16.

1.`FixedHashMap` is ready, which is designed for int8 and int16.
2.Added `CiderNewHashTableTest.cpp` and validated the usage of `FixedHashMap`.
3.Use `CiderDefaultAllocator` as default allocator for memory usage in HashTable.

Implement new HashTables.

No.

UTs.


Deegue:
APIs for codegen will be submitted in another pr.

Trying to add an exclusion file for license check.

It seems something wrong with code-style-check, there is no difference between the original version and the auto revised version.

cider/common/Stopwatch.h
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
```
jikunshang, line 23: necessary file?

cider/common/hashtable/FixedHashMap.h
```diff
@@ -0,0 +1,189 @@
...
+    this->emplace(x, it, inserted);
+    if (inserted)
+      new (&it->getMapped()) Mapped();
+
+    return it->getMapped();
+  }
+};
+
+template <typename Key, typename Mapped, typename Allocator = HashTableAllocator>
+using FixedImplicitZeroHashMap =
```
jikunshang, line 176: `FixedImplicitZeroHashMap` `FixedImplicitZeroHashMapWithCalculatedSize` `FixedHashMap` are all interface you will provide? what's the difference and how to choose. Do we need ut for these basic class and API? Please provide more info via comments.

cider/common/hashtable/FixedHashSet.h
```diff
@@ -0,0 +1,51 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <common/hashtable/FixedHashTable.h>
+
+template <typename Key, typename Allocator = HashTableAllocator>
+class FixedHashSet
```
jikunshang, line 28: do we need this class?

cider/tests/CiderNewHashTableTest.cpp
```diff
@@ -0,0 +1,265 @@
...
+using namespace TestHelpers;
+
+static const std::shared_ptr<CiderAllocator> allocator =
+    std::make_shared<CiderDefaultAllocator>();
+
+class CiderNewHashTableTest : public ::testing::Test {
+ protected:
+  static MockTable* MockTableForTest;
+
+  static void SetUpTestSuite() {
```
jikunshang, line 41: do we need this?

cider/tests/CiderNewHashTableTest.cpp
```diff
@@ -0,0 +1,265 @@
...
+  int64_t getSum() const { return sum; }
+
+  size_t getCount() const { return count; }
+
+  double getAvg() const { return (double)sum / count; }
+
+  bool operator!() const { return !count; }
+};
+
+TEST_F(CiderNewHashTableTest, aggInt8Test) {
```
jikunshang, line 180: I think this test is not verifying hashmap functionality, it's more like an example about how to do aggregation based on this hashmap.

jikunshang:
```
1.FixedHashMap is ready, which is designed for int8 and int16.
2.Added CiderNewHashTableTest.cpp and validated the usage of FixedHashMap.
```
I don't think you have verified all APIs `FixedHashMap` provided. 
And item 3456 are TODOs.

spevenhe:
Can we rename these class adding ck in their name or change the dir name to ck to indicate where they are from? Right now they are using basic names and may be confusing when we migrate more hashtables from other frameworks.

cider/common/Arena.h
```diff
@@ -0,0 +1,122 @@
...
+
+/** Memory pool to append something. For example, short strings.
+ * Usage scenario:
+ * - put lot of strings inside pool, keep their addresses;
+ * - addresses remain valid during lifetime of pool;
+ * - at destruction of pool, all memory is freed;
+ * - memory is allocated and freed by large MemoryChunks;
+ * - freeing parts of data is not possible (but look at ArenaWithFreeLists if you need);
+ */
+class Arena : private boost::noncopyable {
```
jikunshang, line 41: do we need an ArenaAllocator like https://github.com/duckdb/duckdb/blob/master/src/include/duckdb/storage/arena_allocator.hpp

cider/common/Stopwatch.h
Deegue: It will be used in HashTable.h to record time cost of resize(rehash) when debug is enabled. Anyway, let me remove this since it's our first version and add back later.

cider/common/hashtable/FixedHashMap.h
Deegue: Yes, `FixedImplicitZeroHashMap` is for int8 and `FixedImplicitZeroHashMapWithCalculatedSize` is for int16. Let me add some comments here.

cider/tests/CiderNewHashTableTest.cpp
Deegue: It will be used later, let me remove it now and add back in the future.

cider/common/hashtable/FixedHashSet.h
Deegue: Yes, it may be used in `Aggregator` to store all inserted keys.

jikunshang: Please provide some examples for how to use `Aggregator` if necessary

cider/common/Arena.h
Deegue: Something like that but here it will only be used for string. Added some todos here for further implementation.

cider/common/hashtable/FixedHashSet.h
Deegue: Examples will be available in PR for [POAE7-2678](https://jira.devtools.intel.com/browse/POAE7-2678)

cider/tests/CiderNewHashTableTest.cpp
Deegue: Yes, these are examples of using `FixedImplicitZeroHashMap` and `FixedImplicitZeroHashMapWithCalculatedSize`.

Deegue:
> Can we rename these class adding ck in their name or change the dir name to ck to indicate where they are from? Right now they are using basic names and may be confusing when we migrate more hashtables from other frameworks.

I think we don't need to do that since it's not a copy of code but picked useful parts. Later if we integrate other HTs, the first thing is to find out whether the code we wanna pick is logically existed in our repo or not. Meanwhile, we can also refactor the code if it is necessary.

spevenhe:
> > Can we rename these class adding ck in their name or change the dir name to ck to indicate where they are from? Right now they are using basic names and may be confusing when we migrate more hashtables from other frameworks.
> 
> I think we don't need to do that since it's not a copy of code but picked useful parts. Later if we integrate other HTs, the first thing is to find out whether the code we wanna pick is logically existed in our repo or not. Meanwhile, we can also refactor the code if it is necessary.

In this way, we still need to rename the class. Since we will have many hashtables, only the base interface class can be called ```hashtable``` and if other implemented class are also called ```hashtable``` , their name will be confusing.

Deegue:
> > > Can we rename these class adding ck in their name or change the dir name to ck to indicate where they are from? Right now they are using basic names and may be confusing when we migrate more hashtables from other frameworks.
> > 
> > 
> > I think we don't need to do that since it's not a copy of code but picked useful parts. Later if we integrate other HTs, the first thing is to find out whether the code we wanna pick is logically existed in our repo or not. Meanwhile, we can also refactor the code if it is necessary.
> 
> In this way, we still need to rename the class. Since we will have many hashtables, only the base interface class can be called `hashtable` and if other implemented class are also called `hashtable` , their name will be confusing.

`HashTable` is a basic interface class, it's universal and it won't be changed. Upper implementations may have different names like `FixedHashTable`.

spevenhe:
> > > > Can we rename these class adding ck in their name or change the dir name to ck to indicate where they are from? Right now they are using basic names and may be confusing when we migrate more hashtables from other frameworks.
> > > 
> > > 
> > > I think we don't need to do that since it's not a copy of code but picked useful parts. Later if we integrate other HTs, the first thing is to find out whether the code we wanna pick is logically existed in our repo or not. Meanwhile, we can also refactor the code if it is necessary.
> > 
> > 
> > In this way, we still need to rename the class. Since we will have many hashtables, only the base interface class can be called `hashtable` and if other implemented class are also called `hashtable` , their name will be confusing.
> 
> `HashTable` is a basic interface class, it's universal and it won't be changed. Upper implementations may have different names like `FixedHashTable`.

Yes. But in this PR, the ```hashtable```class in ```hashtable.h``` is not the basic interface class we descripted in MIP. It is more like an implementation. So I am just thinking its name is confusing.

Deegue:
> 



> > > > > Can we rename these class adding ck in their name or change the dir name to ck to indicate where they are from? Right now they are using basic names and may be confusing when we migrate more hashtables from other frameworks.
> > > > 
> > > > 
> > > > I think we don't need to do that since it's not a copy of code but picked useful parts. Later if we integrate other HTs, the first thing is to find out whether the code we wanna pick is logically existed in our repo or not. Meanwhile, we can also refactor the code if it is necessary.
> > > 
> > > 
> > > In this way, we still need to rename the class. Since we will have many hashtables, only the base interface class can be called `hashtable` and if other implemented class are also called `hashtable` , their name will be confusing.
> > 
> > 
> > `HashTable` is a basic interface class, it's universal and it won't be changed. Upper implementations may have different names like `FixedHashTable`.
> 
> Yes. But in this PR, the `hashtable`class in `hashtable.h` is not the basic interface class we descripted in MIP. It is more like an implementation. So I am just thinking its name is confusing.

`HashTable.h` is the basic interface class. APIs defined in our MIP changed recently and I have aligned `HashTable.h` to it.

cider/common/Arena.h
```diff
@@ -0,0 +1,122 @@
...
+
+#pragma once
+
+#include <cstring>
+#include <memory>
+#include <vector>
+
+#include <boost/noncopyable.hpp>
+
+namespace DB {
```
chaojun-zhang, line 31: why use DB namespace?

Deegue: Removed ns since it's useless currently.

Deegue:
Gentle ping @jikunshang @spevenhe @chaojun-zhang for another review, thanks!

jikunshang:
Please add some test case for hash table API. I don't think you have any case to cover emplace/find/etc. methods.
BTW, please try to rewrite PR descirption.

Deegue:
> Please add some test case for hash table API. I don't think you have any case to cover emplace/find/etc. methods. BTW, please try to rewrite PR descirption.

Added api test, rewrote change 3 of PR descirption.

src/cider/tests/CiderNewHashTableTest.cpp
```diff
@@ -0,0 +1,216 @@
...
+  for (int i = 0; i < keys.size(); i++) {
+    CHECK_EQ(map[keys[i]].getSum(), 60);
+    CHECK_EQ(map[keys[i]].getCount(), 2);
+    CHECK_EQ(map[keys[i]].getAvg(), 30);
+  }
+}
+
+TEST_F(CiderNewHashTableTest, hashTableTest) {
+  using Cell = FixedHashMapCell<int16_t, Block>;
+  FixedHashTable<int16_t, Cell, FixedHashTableStoredSize<Cell>, HashTableAllocator> fht;
```
jikunshang, line 155: the second parameter have to be `FixedHashMapCell`, possible to be some simple type for this ut/example?

jikunshang (approved):

yma11 (approved):

src/cider/tests/CiderNewHashTableTest.cpp
yma11: @Deegue please address this comment in later PR.

jikunshang:
please add `common` to cpplint tool, and fix potential lint issue. see https://github.com/intel/BDTK/pull/267

src/cider/tests/CiderNewHashTableTest.cpp
Deegue: Users will not need to care about Cell like `FixedHashMapCell<key_type, value_type>` or `HashMapCell<key_type, value_type>` since it's related to the HashTable implementation decided by the chooser. Users only need to pass down `key_type` and `value_type`.

jikunshang: what should I do if I want a simple hash map which `key_type` is Int32 and `value_type` is Int64

Deegue: When using `Aggregator`, you can set via constructor. But for hash map, it cannot avoid setting a `Cell`. It's not universal for simple hash map.

Exposing `Aggregator`, `Join` may be better than simple `hashtable` to users for easier integration.

jikunshang: if so, `Cell` will be part of `hashtable` API, please call out that.

PR #256, opened 2022-12-15
[POAE7-2687] support string comparison and like-expr in nextgen

YBRua:

- string-/varchar-related features in nextgen framework
  - added support for string comparison and string LIKE
  - added support for creating string literals (using LLVM global strings)
  - added support for casting from varchar to varchar (actually this cast does nothing and returns the original varchar)
  - fixed arrow array buffer allocation: initialize newly allocated buffers with 0's to prevent accidentally reading garbage data from the uninitialized buffer
- unit test migration
  - migrated existing string unit tests to nextgen
  -   currently only is-null, comparisons and likes are enabled, other cases are not yet supported

Migration to nextgen framework.

No.

By unit tests.


src/cider/exec/nextgen/parsers/RelAlgEUParser.cpp
```diff
@@ -96,6 +96,9 @@ class InputAnalyzer {
     } else {
       auto children = (*curr)->get_children_reference();
       for (ExprPtr* child : children) {
+        if (!child || !child->get()) {
```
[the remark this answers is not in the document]
YBRua: in like expr there are 3 arguments: the input, the 'like' pattern and an escape character. Currently the escape character is nullptr (because substrait does not support escape)

src/cider/tests/nextgen/functional/CiderStringNextgenTest.cpp
```diff
@@ -0,0 +1,1048 @@
...
+          .build();
+  std::shared_ptr<CiderBatch> res_batch = std::make_shared<CiderBatch>(
+      schema, array, std::make_shared<CiderDefaultAllocator>());
+
+  assertQueryArrow("SELECT col_1, COUNT(*) FROM test GROUP BY col_1", res_batch, true);
+}
+
+TEST_F(CiderDuplicateStringTestNextGen, MultiGroupKeyTest) {
+  // TODO: (YBRua) Enable this after nextgen supports GROUP BY
+  GTEST_SKIP();
```
yma11, line 223: Add message for SKIP reason.

src/cider/tests/nextgen/functional/CiderStringNextgenTest.cpp
```diff
@@ -0,0 +1,1048 @@
...
+  //                             .build();
+  // assertQuery("SELECT col_1 FROM test where col_1 LIKE '[!aa]%'",
+  // expected_batch_3);
+}
+
+// stringop: substring
+
+TEST_F(CiderStringNullableTestNextGen, SubstringTest) {
+  // TODO: (YBRua) Enable this after nextgen supports StringOp
+  GTEST_SKIP();
```
yma11, line 292: Same as above.

src/cider/type/plan/UnaryExpr.cpp
```diff
@@ -37,36 +37,83 @@ JITExprValue& UOper::codegen(JITFunction& func) {
...
+      return codegenCast(func, operand);
     }
     default:
       UNIMPLEMENTED();
   }
 }
 
+JITExprValue& UOper::codegenIsNull(JITFunction& func,
+                                   Analyzer::Expr* operand,
+                                   bool use_is_not_null) {
```
yma11, line 59: `use_is_not_null` is a little tricky here, why not directly pass the `get_optype()` here?

src/cider/type/plan/UnaryExpr.cpp
```diff
@@ -37,36 +37,83 @@ JITExprValue& UOper::codegen(JITFunction& func) {
...
+  const auto& ti = get_type_info();
+  const auto type = ti.is_decimal() ? decimal_to_int_type(ti) : ti.get_type();
+
+  // for IS NULL / IS NOT NULL, we only need the null info (getNull())
+  JITExprValueAdaptor operand_val(operand->codegen(func));
+
+  if (use_is_not_null) {
+    // is not null
+    auto null_value = operand_val.getNull().get()
+                          ? operand_val.getNull()->notOp()
```
yma11, line 70: seems `operand_val.getNull().get()` always exists even for non-nullable columns. Logic here need updated.`null_value=operand_val.getNull()->notOp()`.

src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITUtils.h
```diff
@@ -99,6 +99,15 @@ inline llvm::Value* getLLVMConstantFP(double value,
...
   }
 }
+
+inline llvm::Value* getLLVMConstantGlobalStr(const std::string& value,
+                                             llvm::IRBuilder<>* builder,
+                                             llvm::LLVMContext& ctx) {
+  llvm::Value* str_lv = builder->CreateGlobalString(
+      value, "str_const_" + std::to_string(std::hash<std::string>()(value)));
+  str_lv = builder->CreateBitCast(str_lv, getLLVMPtrType(JITTypeTag::INT8, ctx));
+  return str_lv;
```
jikunshang, line 109: Do we need length? or it ends with `\0`.
Or we always need to get length info from expression since it's a constant expression.

jikunshang (commented):
Over LGTM.

src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITUtils.h
YBRua: The value returned here does not contain length. If length is needed, it can be acquired from the `std::string` input (for C++), or from the `JITExprValue` of the returned constant expression (for LLVM IR).

BTW, seems the global string is nul-terminated, [doc](https://llvm.org/doxygen/classllvm_1_1IRBuilderBase.html#a3a7e4483b56895f73ad5898251dc4b4c).

src/cider/type/plan/LikeExpr.h
```diff
@@ -0,0 +1,125 @@
...
+                              is_ilike,
+                              is_simple);
+  }
+  bool operator==(const Expr& rhs) const override;
+  std::string toString() const override;
+  void find_expr(bool (*f)(const Expr*),
+                 std::list<const Expr*>& expr_list) const override;
+  std::shared_ptr<Analyzer::Expr> get_shared_arg() { return arg; }
+  std::shared_ptr<Analyzer::Expr> get_shared_Like() { return like_expr; }
+  std::shared_ptr<Analyzer::Expr> get_shared_escape() { return escape_expr; }
```
yma11, line 105: Can you add comment on the methods which can be removed in nextgen in future?

yma11 (approved):

jikunshang (approved):

PR #258, opened 2022-12-15
[POAE7-2702] Updata in function yaml-file to support in:str_list

haiweizh:

Updata in function yaml-file and code to support in:str_list
Update velox test case code to support function mapping between presto and cider.

Updata in function yaml-file to support in:str_list

No

Uts

cider

### DIFF
--- a/src/cider/function/internals/functions_string.yaml
+++ b/src/cider/function/internals/functions_string.yaml
@@ -1511,6 +1511,59 @@ scalar_functions:
           - value: i64
             name: "group"
         return: "string"
+  - 
+    name: "in"
+    description: >-
+      Whether the `expression` is in value list.
+    impls:
+      - args:
+          - value: string
+            name: expression
+            description: The expression to test for in the value list.
+          - value: string
+            name: list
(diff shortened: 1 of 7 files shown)


src/cider/function/internals/functions_string.yaml
```diff
@@ -1511,6 +1511,59 @@ scalar_functions:
           - value: i64
             name: "group"
         return: "string"
+  - 
+    name: "in"
```
winningsix, line 1515: Q: Does this mean we need to maintain sth similar to the upstream yaml moving forward? https://github.com/substrait-io/substrait/commit/81e34d4054ff0dbde23ac749fbb8fcc232989c5d#diff-b05c613f1c6a478bd62cd87456ade235540725c12afced21e5c3d534cd1da1e5R748

haiweizh: Yes, we need to update the yaml file when cider add new feature.

jikunshang (approved):

yma11 (approved):

PR #263, opened 2022-12-16
[POAE7-2713] Enable and support filter TCs in nextgen

yma11:
1. add null_vector for non-nullable cols
2. fill null_count for output arrow array
3. port filter TCs

nextgen support

No

UT

### DIFF
--- a/src/cider/exec/nextgen/context/RuntimeContext.h
+++ b/src/cider/exec/nextgen/context/RuntimeContext.h
@@ -53,8 +53,8 @@ class RuntimeContext {
     auto length = arrow_array->length;
     auto arrow_schema = batch->getSchema();
 
-    auto process_function =
-        utils::RecursiveFunctor{[&length](auto&& process_function,
+    auto set_null_count_function =
+        utils::RecursiveFunctor{[&length](auto&& set_null_count_function,
                                           ArrowArray* arrow_array,
                                           ArrowSchema* arrow_schema) -> void {
           if (arrow_array->buffers[0]) {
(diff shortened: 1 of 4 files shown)


yma11:
@jikunshang @bigPYJ1151 please help take a review.

src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -186,6 +154,38 @@ class ColumnWriter {
     }
   }
 
+  JITValuePointer setNullBuffer(JITValuePointer& null_val) {
```
jikunshang, line 157: rename `setNullBuffer` ?

src/cider/exec/nextgen/context/RuntimeContext.h
```diff
@@ -46,7 +48,27 @@ class RuntimeContext {
     if (batch_holder_.empty()) {
       return nullptr;
     }
-    return batch_holder_.back().second.get();
+    auto batch = batch_holder_.back().second.get();
+    auto arrow_array = batch->getArray();
+    auto length = arrow_array->length;
+    auto arrow_schema = batch->getSchema();
+
+    auto process_function =
```
jikunshang, line 56: rename to `set_null_count_function`?

jikunshang (approved):

src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -186,6 +154,38 @@ class ColumnWriter {
...
+
+      context_.getJITFunction()->emitRuntimeFunctionCall(
+          "set_null_vector_bit",
+          JITFunctionEmitDescriptor{
+              .ret_type = JITTypeTag::VOID,
+              .params_vector = {{null_buffer.get(),
+                                 index_.get(),
+                                 context_.getJITFunction()
+                                     ->createConstant(JITTypeTag::BOOL, false)
+                                     .get()}}});
```
bigPYJ1151, line 184: Hmmm, I think it's better to write a memset runtime function and set null buffer with 0 in local variable block, rather than set it bit-by-bit inside the process loop, which is inefficient. BTW, I forget why all output columns should be allocated a null buffer?😂 Any cases will break if we don't do it?

bigPYJ1151 (approved):

jikunshang (approved):

PR #265, opened 2022-12-19
[POAE7-2725] NextGen segfaults if no column is referenced in input query

YBRua:
#### Bug Description

If an input query did not reference any column in the input table, e.g. `select 'a-string-literal' from test` or `select 2 from test`, a segmentation fault would occur during ColumnToRow codegen.

This is because, `ColumnToRowNode` gets input from preceding `ArrowSourceNode`. `ArrowSourceNode` keeps all columns that will be used in a vector `input_exprs_`. When inserting `ArrowSourceNode` into the OpNodePipeline, an `InputAnalyzer` will remove any column that is not used in the query (Around L71 in RelAlgEUParser.cpp). In the example queries above, no column is referenced, so all columns are erased from `input_exprs_`. As a result, `input_exprs_` is an *empty vector*.

However, the `ColumnToRowNode` attempts to access the first element in the `input_exprs_` in order to get the row number of the input table (Around L152 in ColumnToRowNode.cpp). Since `input_exprs_` is empty, it results in segmentation fault.

#### Fix

Fixed by adding a special-case handling in `ColumnToRowNode`, if `input_exprs_` is empty, the node tries to get the length directly from the input arrow array (which is available as a parameter passed to the query function).

---

@bigPYJ1151 please help take a review, thanks.

Bug fix. Cause of the bug and fixes are detailed in the above section.

No.

UT. Added a unit test in FilterProjectTest.

BUG

### DIFF
--- a/src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
+++ b/src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
@@ -150,10 +150,22 @@ void ColumnToRowTranslator::codegen(context::CodegenContext& context) {
   auto index = func->createVariable(JITTypeTag::INT64, "index", 0);
 
   // input column rows num.
-  auto len = func->createLocalJITValue([&context, &inputs]() {
-    return context::codegen_utils::getArrowArrayLength(
-        context.getArrowArrayValues(inputs.front()->getLocalIndex()).first);
-  });
+  JITValuePointer len = nullptr;
+  if (inputs.size()) {
+    len.replace(func->createLocalJITValue([&context, &inputs]() {
+      return context::codegen_utils::getArrowArrayLength(
+          context.getArrowArrayValues(inputs.front()->getLocalIndex()).first);
+    }));
+  } else {
+    // inputs will be empty if no column is referenced in the sql query
+    // e.g. select 123 from test, so we directly get length from the input arrow array
+    // assumes signature be like: query_func(context, array)
+    auto input_array = func->getArgument(1);  // array
+    len.replace(func->createLocalJITValue([&context, &input_array]() {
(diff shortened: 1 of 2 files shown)


src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
```diff
@@ -150,10 +150,22 @@ void ColumnToRowTranslator::codegen(context::CodegenContext& context) {
...
+  if (inputs.size()) {
+    len.replace(func->createLocalJITValue([&context, &inputs]() {
+      return context::codegen_utils::getArrowArrayLength(
+          context.getArrowArrayValues(inputs.front()->getLocalIndex()).first);
+    }));
+  } else {
+    // inputs will be empty if no column is referenced in the sql query
+    // e.g. select 123 from test, so we directly get length from the input arrow array
+    // assumes signature be like: query_func(context, array)
+    auto input_array = func->getArgument(1);  // array
```
yma11, line 163: maybe we can get len from `input_array` even `inputs` is not empty?

YBRua: Yes, but I left it as is because I thought there was some reasons for this (?)

yma11: @bigPYJ1151 can you take a look at this?

YBRua: updated

bigPYJ1151 (dismissed):

jikunshang (dismissed):

yma11 (approved):

jikunshang (approved):

PR #266, opened 2022-12-19
[POAE7-2688] String op workflow in nextgen and substring impl.

jikunshang:

Add string op workflow for nextgen compiler, and add substring function impl.

No.

UT


jikunshang:
@bigPYJ1151 @YBRua @yma11 Please take a review, thanks.

src/cider/type/plan/BinaryExpr.cpp
```diff
@@ -39,7 +39,9 @@ JITExprValue& BinOper::codegen(JITFunction& func) {
 
   const auto& lhs_ti = lhs->get_type_info();
   const auto& rhs_ti = rhs->get_type_info();
-  CHECK_EQ(lhs_ti.get_type(), rhs_ti.get_type());
+  if (!lhs_ti.is_string()) {
```
YBRua, line 42: I think, (though it is not very important), for completeness and type safety, when `lhs_ti` is a string type, perhaps we should type-check that `rhs_ti` is also a string type?

src/cider/function/scalar/RuntimeFunctions.cpp
```diff
@@ -1075,6 +1075,7 @@ extern "C" ALWAYS_INLINE uint64_t string_pack(const int8_t* ptr, const int32_t l
 }
 
 #ifdef __clang__
+#include "function/string/CiderStringFunctions.cpp"
```
yma11, line 1078: Better to put these new functions under nextgen/function folder?

src/cider/exec/nextgen/context/RuntimeContext.h
```diff
@@ -29,12 +29,16 @@
 namespace cider::exec::nextgen::context {
 class RuntimeContext {
  public:
-  RuntimeContext(int64_t ctx_num) : runtime_ctx_pointers_(ctx_num, nullptr) {}
+  RuntimeContext(int64_t ctx_num) : runtime_ctx_pointers_(ctx_num, nullptr) {
+    string_heap_ptr_ = std::make_shared<StringHeap>();
```
yma11, line 33: Maybe the string heap should initialize during codegen if need? like call `runtime_ctx.setStringHeap()`.

src/cider/function/scalar/RuntimeFunctions.cpp
jikunshang: updated.

src/cider/exec/nextgen/context/RuntimeContext.h
jikunshang: make sense, updated. Please take a look again.

bigPYJ1151 (approved):

yma11 (approved):

PR #270, opened 2022-12-20
[POAE7-2679] CodegenContext Register Hashtable

qiuyangshen:

- provide registerHashTable related function in codegenContext and setHashTable in RuntimeContext
- add related test file
- **(Since the design of the template class used by hashtable is not applicable this time, a hard-coded template class is now used, and the default template class will be provided later, and corresponding modifications will be made here)**

for NextGen FW and HashJoinTranslator development

No

UTs

INFRA

### DIFF
--- a/src/cider/exec/nextgen/CMakeLists.txt
+++ b/src/cider/exec/nextgen/CMakeLists.txt
@@ -35,4 +35,5 @@ add_library(
   nextgen STATIC
   Nextgen.cpp $<TARGET_OBJECTS:cider_operators> $<TARGET_OBJECTS:cider_context>
   $<TARGET_OBJECTS:cider_parsers> $<TARGET_OBJECTS:cider_transformer>)
-target_link_libraries(nextgen ${NEXTGEN_DEPS} jitlib)
+# TODO(qiuyang) : folly will deleted after hashtable refactoring
+target_link_libraries(nextgen ${NEXTGEN_DEPS} jitlib folly)
--- a/src/cider/exec/nextgen/context/CodegenContext.cpp
+++ b/src/cider/exec/nextgen/context/CodegenContext.cpp
@@ -109,6 +109,27 @@ JITValuePointer CodegenContext::registerBuffer(const int32_t capacity,
   return ret;
 }
 
+JITValuePointer CodegenContext::registerHashTable(const std::string& name) {
+  int64_t id = acquireContextID();
+  auto index = this->jit_func_->createLiteral(JITTypeTag::INT64, id);
+  JITValuePointer ret = jit_func_->createLocalJITValue([this, id]() {
+    auto index = this->jit_func_->createLiteral(JITTypeTag::INT64, id);
+    auto pointer = this->jit_func_->emitRuntimeFunctionCall(
+        "get_query_context_item_ptr",
+        JITFunctionEmitDescriptor{
+            .ret_type = JITTypeTag::POINTER,
+            .ret_sub_type = JITTypeTag::INT8,
+            .params_vector = {this->jit_func_->getArgument(0).get(), index.get()}});
+
+    return pointer;
+  });
+  ret->setName(name);
+
+  hashtable_descriptor_.first = std::make_shared<HashTableDescriptor>(id, name);
+  hashtable_descriptor_.second.replace(ret);
+  return ret;
+}
+
 RuntimeCtxPtr CodegenContext::generateRuntimeCTX(
     const CiderAllocatorPtr& allocator) const {
   auto runtime_ctx = std::make_unique<RuntimeContext>(getNextContextID());
@@ -120,6 +141,8 @@ RuntimeCtxPtr CodegenContext::generateRuntimeCTX(
   for (auto& buffer_desc : buffer_descriptors_) {
     runtime_ctx->addBuffer(buffer_desc.first);
   }
+
+  runtime_ctx->addHashTable(hashtable_descriptor_.first);
 
   runtime_ctx->instantiate(allocator);
   return runtime_ctx;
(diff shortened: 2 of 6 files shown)


src/cider/tests/nextgen/compiler/ContextTest.cpp
```diff
@@ -120,6 +125,89 @@ TEST_F(ContextTests, ContextBufferTest) {
...
+
+  query_func((int8_t*)runtime_ctx.get());
+
+  EXPECT_EQ(runtime_ctx->getContextItemNum(), 1);
+  EXPECT_NE(runtime_ctx->getContextItem(0), nullptr);
+
+  auto hash_table = reinterpret_cast<
+      cider_hashtable::
+          LinearProbeHashTable<int, std::pair<Batch*, int>, murmurHash, Equal>*>(
+      runtime_ctx->getContextItem(0));
```
jikunshang, line 186: why not provide a separate `getHashTablePtr` method instead? if you do register batch and register hashtable in codegen workflow, how to get correct index?

src/cider/exec/nextgen/context/CodegenContext.h
```diff
@@ -92,16 +114,41 @@ class CodegenContext {
         : ctx_id(id), name(n), capacity(c), initializer_(initializer) {}
   };
 
+  struct HashTableDescriptor {
+    int64_t ctx_id;
+    std::string name;
+    // TODO(qiuyang) : LinearProbeHashTable will provide the default template
+    LinearProbeHashTable<int, std::pair<Batch*, int64_t>, murmurHash, Equal>* hash_table;
```
jikunshang, line 121: I suggest to define and use base class pointer here. Otherwise we may have huge change(or redesign) in the future. cc @spevenhe

src/cider/exec/nextgen/CMakeLists.txt
```diff
@@ -35,4 +35,4 @@ add_library(
   nextgen STATIC
   Nextgen.cpp $<TARGET_OBJECTS:cider_operators> $<TARGET_OBJECTS:cider_context>
   $<TARGET_OBJECTS:cider_parsers> $<TARGET_OBJECTS:cider_transformer>)
-target_link_libraries(nextgen ${NEXTGEN_DEPS} jitlib)
+target_link_libraries(nextgen ${NEXTGEN_DEPS} jitlib folly)
```
jikunshang, line 38: why need folly? maybe link hashtable lib would be enough?

src/cider/exec/nextgen/context/CodegenContext.cpp
```diff
@@ -84,6 +84,26 @@ JITValuePointer CodegenContext::registerBuffer(const int32_t capacity,
   return ret;
 }
 
+JITValuePointer CodegenContext::registerHashTable(const std::string& name) {
+  int64_t id = acquireContextID();
+  auto index = this->jit_func_->createLiteral(JITTypeTag::INT64, id);
+  JITValuePointer ret = jit_func_->createLocalJITValue([this, id]() {
+    auto pointer = this->jit_func_->emitRuntimeFunctionCall(
+        "get_query_context_ptr",
```
yma11, line 92: Where is this function defined?

src/cider/tests/nextgen/compiler/ContextTest.cpp
qiuyangshen: This is mainly to ensure the consistency of the interface，like registerBatch and registerBuffer. and to get correct index, we have a acquireContextID() function above, This method is an auto-incrementer to make sure it doesn't repeat，and will save the only id in variable hashtable_descriptor_.

yma11: I think the question is, you register this hashtable during translator and assign it an index, but during codegen, how do we know the correct index if there are multi hastables?

yma11: Do you consider multi join nodes case? There will be multi `HashTableDescriptorPtr` and each one connects to a hash table. But how do we know which one we should use(the correct index) during codegen?

yma11: Synced offline, seems multi join nodes are not considered now which means we assume that there is only one hash table for join.

qiuyangshen: Yes, I think even in this case, just replace hashtable with vector<hashtable>,  register multiple times can also guarantee the uniqueness of its context_id by the time

src/cider/exec/nextgen/context/RuntimeContext.h
```diff
@@ -78,6 +80,7 @@ class RuntimeContext {
   std::vector<std::pair<CodegenContext::BatchDescriptorPtr, BatchPtr>> batch_holder_;
   std::vector<std::pair<CodegenContext::BufferDescriptorPtr, BufferPtr>> buffer_holder_;
   std::shared_ptr<StringHeap> string_heap_ptr_;
+  CodegenContext::HashTableDescriptorPtr hashtable_holder_;
```
yma11, line 83: Shouldn't the `hashtable_holder_` hold a pair with descriptor and hash_table ptr, similar like `batch_holder_`? @bigPYJ1151

src/cider/exec/nextgen/CMakeLists.txt
Ma-Jian1: folly is too heavy， please try to avoid using third-party lib in nextgen

spevenhe: hashtable's impl will move outside h file to cpp file in future PR and will be a library. So nextgen will not use folly by then

xieqi (approved):

PR #277, opened 2022-12-22
[POAE7-2665] HashJoinTranslator support 

qiuyangshen:

- Add HashJoinNode and Translator (inner hashjoin) and related RuntimeFuncitons, modify the parser to create HashJoinNode and add it to transform pipeline
- Add HashJoinTest, Construct a basic hash join sql to run the workflow to prove correctness
- Add null vector handling
- Refactor EU parser
- Change the createConstant using to createLiteral in some places
- Remove folly dependency
- Rebase hashtable  .h and .cpp file separation #304 

For HashJoin codegen in Next Gen FW design

No

UTs

INFRA

### DIFF
--- a/src/cider/exec/nextgen/Nextgen.cpp
+++ b/src/cider/exec/nextgen/Nextgen.cpp
@@ -28,7 +28,7 @@
 namespace cider::exec::nextgen {
 
 std::unique_ptr<context::CodegenContext> compile(
-    const RelAlgExecutionUnit& ra_exe_unit,
+    RelAlgExecutionUnit& ra_exe_unit,
     const jitlib::CompilationOptions& co,
     const context::CodegenOptions& codegen_options) {
   auto codegen_ctx = std::make_unique<context::CodegenContext>();
--- a/src/cider/exec/nextgen/context/Buffer.h
+++ b/src/cider/exec/nextgen/context/Buffer.h
@@ -42,11 +42,10 @@ class Buffer {
   void allocateBuffer(int32_t size) {
     if (capacity_) {
       buffer_ = allocator_->reallocate(buffer_, capacity_, size);
-      capacity_ = size;
     } else {
       buffer_ = allocator_->allocate(size);
-      capacity_ = size;
     }
+    capacity_ = size;
   }
 
   int8_t* getBuffer() { return buffer_; }
--- a/src/cider/exec/nextgen/context/Buffer.h
+++ b/src/cider/exec/nextgen/context/Buffer.h
@@ -39,6 +39,16 @@ class Buffer {
 
   ~Buffer() { allocator_->deallocate(buffer_, capacity_); }
 
+  void allocateBuffer(int32_t size) {
+    if (capacity_) {
+      buffer_ = allocator_->reallocate(buffer_, capacity_, size);
+      capacity_ = size;
+    } else {
+      buffer_ = allocator_->allocate(size);
+      capacity_ = size;
+    }
+  }
+
   int8_t* getBuffer() { return buffer_; }
 
   int32_t getCapacity() { return capacity_; }
--- a/src/cider/exec/nextgen/operators/HashJoinNode.cpp
+++ b/src/cider/exec/nextgen/operators/HashJoinNode.cpp
@@ -152,6 +152,7 @@ void traverse(ExprPtr expr,
               std::vector<JITValuePointer>& nulls,
               context::CodegenContext& context) {
   if (Analyzer::ColumnVar* col_var = dynamic_cast<Analyzer::ColumnVar*>(expr.get())) {
+    // FIXME (qiuyang):: 100 is not always used as right table id.
     if (col_var->get_table_id() == 100) {
       utils::FixSizeJITExprValue jit_value(expr->codegen(context));
       keys.emplace_back(jit_value.getValue());
--- a/src/cider/exec/nextgen/operators/HashJoinNode.cpp
+++ b/src/cider/exec/nextgen/operators/HashJoinNode.cpp
@@ -145,20 +145,25 @@ void HashJoinTranslator::consume(context::CodegenContext& context) {
   codegen(context);
 }
 
-// traverse join_quals expr tree to get join key value
+// traverse join_quals expr tree to get join key value and null vector
 void traverse(ExprPtr expr,
               jitlib::JITFunctionPointer func,
-              std::vector<JITValuePointer>& params) {
+              std::vector<JITValuePointer>& keys,
+              std::vector<JITValuePointer>& nulls,
+              context::CodegenContext& context) {
   if (Analyzer::ColumnVar* col_var = dynamic_cast<Analyzer::ColumnVar*>(expr.get())) {
     if (col_var->get_table_id() == 100) {
-      utils::FixSizeJITExprValue jit_value(col_var->codegen(*func));
-      // TODO(qiuyang): null vector handle
-      params.emplace_back(jit_value.getValue());
+      utils::FixSizeJITExprValue jit_value(expr->codegen(context));
+      keys.emplace_back(jit_value.getValue());
+      // null vector handle
+      if (!expr->get_type_info().get_notnull()) {
+        nulls.emplace_back(jit_value.getNull());
+      }
     }
   } else {
     auto children = expr->get_children_reference();
     for (auto child : children) {
-      traverse(*child, func, params);
+      traverse(*child, func, keys, nulls, context);
     }
   }
 }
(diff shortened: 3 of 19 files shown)


src/cider/exec/nextgen/Nextgen.cpp
```diff
@@ -27,7 +27,7 @@
 
 namespace cider::exec::nextgen {
 
-std::unique_ptr<context::CodegenContext> compile(const RelAlgExecutionUnit& ra_exe_unit,
+std::unique_ptr<context::CodegenContext> compile(RelAlgExecutionUnit& ra_exe_unit,
```
yma11, line 30: Why remove `const`? do we need to change ra_exe_unit during codegen?

qiuyangshen: I used RelAlgExecutionUnit as a member variable in class InputAnalyzer, and traverse access to read and write members of it

src/cider/exec/nextgen/operators/HashJoinNode.cpp
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// traverse join_quals expr tree to get join key value and null vector
+void traverse(ExprPtr expr,
+              jitlib::JITFunctionPointer func,
+              std::vector<JITValuePointer>& keys,
+              std::vector<JITValuePointer>& nulls) {
+  if (Analyzer::ColumnVar* col_var = dynamic_cast<Analyzer::ColumnVar*>(expr.get())) {
+    if (col_var->get_table_id() == 100) {
+      utils::FixSizeJITExprValue jit_value(col_var->codegen(*func));
```
yma11, line 155: Seems you need a rebase since `codegen` API is changed.

src/cider/exec/nextgen/operators/HashJoinNode.cpp
```diff
@@ -0,0 +1,253 @@
...
+  auto join_res_len = func->emitRuntimeFunctionCall(
+      "look_up_value_by_key",
+      JITFunctionEmitDescriptor{
+          .ret_type = JITTypeTag::INT64,
+          .params_vector = {
+              hashtable.get(), key_value.get(), key_null.get(), join_res_buffer.get()}});
+
+  auto build_table_map = dynamic_cast<HashJoinNode*>(node_.get())->getBuildTableMap();
+  auto row_index = func->createVariable(JITTypeTag::INT64, "row_index", 0l);
+  row_index = func->createLiteral(JITTypeTag::INT64, 0l);
```
yma11, line 201: `row_index` should can be created as `Literal` directly.

src/cider/exec/nextgen/parsers/RelAlgEUParser.cpp
```diff
@@ -47,23 +44,40 @@ static bool isParseable(const RelAlgExecutionUnit& eu) {
...
+    if (!eu_.quals.empty()) {
+      for (auto& expr : eu_.quals) {
+        traverse(&expr);
+      }
     }
 
-    for (size_t i = 0; i < input_desc_.size(); ++i) {
-      input_desc_to_index_.insert({input_desc_[i], i});
+    if (!eu_.join_quals.empty()) {
+      // build table always start from index/2 in input_col_descs
```
yma11, line 70: Why?

src/cider/exec/nextgen/operators/HashJoinNode.cpp
qiuyangshen: I'm affraid we can't. I use it as variable in the logic behind , otherwise it will call "JITValue is not a variable in LLVMJITValue::assign." error

src/cider/exec/nextgen/parsers/RelAlgEUParser.cpp
qiuyangshen: The code is obsolete , have changed it.

src/cider/exec/nextgen/operators/HashJoinNode.cpp
qiuyangshen: Roger that.

src/cider/exec/nextgen/operators/HashJoinNode.cpp
yma11: @bigPYJ1151 is this expected?

bigPYJ1151: Yes, only two types of ```JITValue``` could be assigned:
- variable.
- dereference of ```Pointer``` type ```JITValue```.

src/cider/exec/nextgen/Nextgen.cpp
```diff
@@ -28,7 +28,7 @@
 namespace cider::exec::nextgen {
 
 std::unique_ptr<context::CodegenContext> compile(
-    const RelAlgExecutionUnit& ra_exe_unit,
+    RelAlgExecutionUnit& ra_exe_unit,
```
winningsix, line 31: Q: why need to remove ```const``` here? Same for 
https://github.com/intel/BDTK/pull/277/files#diff-e19320dad5e5a53117976e0a305558dae44f37af90b5f7ddf23929825f5c2f57R33

src/cider/exec/nextgen/context/Buffer.h
```diff
@@ -39,6 +39,16 @@ class Buffer {
 
   ~Buffer() { allocator_->deallocate(buffer_, capacity_); }
 
+  void allocateBuffer(int32_t size) {
+    if (capacity_) {
+      buffer_ = allocator_->reallocate(buffer_, capacity_, size);
+      capacity_ = size;
+    } else {
+      buffer_ = allocator_->allocate(size);
+      capacity_ = size;
```
winningsix, line 48: Minor: merge & move L45 & L48 in common path.

src/cider/exec/nextgen/operators/HashJoinNode.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+class BuildTableReader {
+ public:
+  BuildTableReader(utils::JITExprValue& buffer_values,
+                   ExprPtr& expr,
+                   JITValuePointer& index)
+      : buffer_values_(buffer_values), expr_(expr), index_(index) {}
+
+  void read() {
+    switch (expr_->get_type_info().get_type()) {
```
winningsix, line 40: Do we need to check type every time? Or we can get a type inferenced reader?

src/cider/exec/nextgen/Nextgen.cpp
qiuyangshen: I used the RelAlgExecutionUnit in class InputAnalyzer, run the traverse function with its member variable such as join_quals, shared_target_exprs etc. and needs to read and write it in traverse logic
https://github.com/intel/BDTK/pull/277/files#diff-bd9f1527faedcddf56a8c76d328735d0934369763a3e50a4315fc346df8b63c9R109

src/cider/exec/nextgen/context/Buffer.h
qiuyangshen: FIX

src/cider/exec/nextgen/operators/HashJoinNode.cpp
qiuyangshen: One row and multiple columns in the build table are traversed each time here, and the expr of different columns requires different processing logic. The whole logic is in a loop, and this level of abstraction is consistent with the logic of ColumnToRow Translator‘s ColumnReader.

yma11: I think it the codegen logic and once codegen done, no type check will need in row function right?

src/cider/exec/nextgen/operators/HashJoinNode.cpp
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// traverse join_quals expr tree to get join key value and null vector
+void traverse(ExprPtr expr,
+              jitlib::JITFunctionPointer func,
+              std::vector<JITValuePointer>& keys,
+              std::vector<JITValuePointer>& nulls,
+              context::CodegenContext& context) {
+  if (Analyzer::ColumnVar* col_var = dynamic_cast<Analyzer::ColumnVar*>(expr.get())) {
+    if (col_var->get_table_id() == 100) {
```
yma11, line 155: I think you'd better add a FIXME here since `100` is not always used as right table id.

yma11 (approved):

src/cider/exec/nextgen/operators/HashJoinNode.cpp
qiuyangshen: FIX

src/cider/exec/nextgen/operators/HashJoinNode.cpp
qiuyangshen: Yes, the different column types of each row are consistent, but here are the different columns of a row, and the respective types may be different.

yma11 (approved):

xieqi (approved):

PR #279, opened 2022-12-23
[POAE7-2747][perf] cache compiled function ptr for performance boost

Ma-Jian1:
cache compiled function ptr for performance boost

performance

No

no need

INFRA


Ma-Jian1:
cache function pointer in batchprocessor

PR #281, opened 2022-12-23
[POAE7-2749] [Hotfix] Fix Wrong Function Declaration Copy Procedure in cloneFunctionRecursive()

bigPYJ1151:

Fix Wrong Function Declaration Copy Procedure in cloneFunctionRecursive()

BugFix

BugFix

Pls take a review @kd97 @yma11

### DIFF
--- a/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
+++ b/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
@@ -222,7 +222,8 @@ JITValuePointer LLVMJITFunction::emitRuntimeFunctionCall(
 
 void LLVMJITFunction::cloneFunctionRecursive(llvm::Function* fn) {
   CHECK(fn);
-  if (fn->isDeclaration()) {
+  if (!fn->isDeclaration()) {
+    // Function has been cloned.
     return;
   }
   // Get the implementation from the runtime module.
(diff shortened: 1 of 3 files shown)


yma11 (approved):
I verified it in my scenario and it works. But logic is weird here and seems we actually disable the recursive clone since in most cases `fn->isDeclaration()` is true?

bigPYJ1151:
> I verified it in my scenario and it works. But logic is weird here and seems we actually disable the recursive clone since in most cases `fn->isDeclaration()` is true?

Emm, you're right, seems it not the root cause.

kd97 (dismissed):

bigPYJ1151:
Pls help to check, thx. @yma11 @kd97

src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
```diff
@@ -254,7 +255,11 @@ void LLVMJITFunction::cloneFunctionRecursive(llvm::Function* fn) {
   for (auto it = llvm::inst_begin(fn), e = llvm::inst_end(fn); it != e; ++it) {
     if (llvm::isa<llvm::CallInst>(*it)) {
       auto& call = llvm::cast<llvm::CallInst>(*it);
-      cloneFunctionRecursive(call.getCalledFunction());
+      llvm::Function* called_func = call.getCalledFunction();
+      if (called_func && called_func->hasFnAttribute(llvm::Attribute::AlwaysInline)) {
+        // Only symbol-called function with always-inline atrribute needs to clone.
+        cloneFunctionRecursive(called_func);
+      }
     }
```
kd97, line 263: what determines the `AlwaysInline` attribute (eg, func definition) ? And no need to clone if not go through this branch ?

kd97: seems AlwaysInline equals recursion termination condition？

bigPYJ1151: Runtime function defined with ```ALWAYSINLINE``` has ```AlwaysInline``` attribute. The only reason to clone function definition is inlining the function, or function declaration is enough.

yma11: So for the runtime function which called object's API should not use `ALWAYS_INLINE` label? like call `ordered_set::insert()`.

bigPYJ1151: > So for the runtime function which called object's API should not use `ALWAYS_INLINE` label? like call `ordered_set::insert()`.

No such restriction.

yma11 (approved):

kd97 (approved):

PR #280, opened 2022-12-23
[POAE7-2718] Migrate MathOp tests to nextgen

imajiayu:
### What changes were proposed in this pull request?
support div zero&overflow error check for nextgen

### Does this PR introduce _any_ user-facing change?
exceptions will be thrown if any error is detected. 

### How was this patch tested?
tests/nextgen/functional/CiderMathOpNextgenTest.cpp

PR #285, opened 2022-12-26
[POAE7-2742] Concat & Trim stringops support in NextGen

YBRua:

Added support for stringop CONCAT and TRIM in nextgen framework

- Added support for string concatenation
  - Added dynamic string data buffer re-allocation. The logic is the same as is in template-based codegen: data buffer will be re-allocated whenever more than 90% of its capacity is used.
  - String concatenation now supports concatenating two columns.
- Added support for string trimming
  - Added a trim map in codegen and runtime context. At compile time, for each `TrimStringOper`, characters to be trimmed will be registered and stored in codegen context. At runtime, the characters to be trimmed will be passed to runtime functions to perform trimming.
  - Changed interfaces of `Expr::codegen()` so that expr-level codegen can have access to codegen context.
- Enabled related UTs.

String-related nextgen UT migration and feature enablement.

No.

UTs.


src/cider/type/plan/Expr.cpp
```diff
@@ -30,6 +30,10 @@ JITExprValue& Expr::codegen(JITFunction& func) {
   UNREACHABLE();
   return expr_var_;
 }
+
+JITExprValue& Expr::codegen(JITFunction& func, context::CodegenContext& context) {
```
jikunshang, line 34: please see this discussion thread https://github.com/intel/BDTK/pull/280#discussion_r1057155512

jikunshang (dismissed):

jikunshang (approved):

src/cider/exec/nextgen/context/CodegenContext.cpp
```diff
@@ -169,10 +169,26 @@ RuntimeCtxPtr CodegenContext::generateRuntimeCTX(
...
   return runtime_ctx;
 }
 
+int CodegenContext::registerTrimStringOperCharMap(const std::string& trim_chars) {
+  if (!trim_char_maps_) {
+    trim_char_maps_ = std::make_shared<std::vector<std::vector<int8_t>>>();
+  }
+
+  auto trim_char_map = std::vector<int8_t>(256, 0);
+  // initialize map and set characters to be trimmed to 1
```
kd97, line 184: better to use map here?

src/cider/exec/nextgen/context/RuntimeContext.cpp
```diff
@@ -125,4 +125,13 @@ Batch* RuntimeContext::getNonGroupByAggOutputBatch() {
...
 }
 
+void RuntimeContext::setTrimStringOperCharMaps(
+    const CodegenContext::TrimCharMapsPtr& maps) {
+  trim_char_maps_ = maps;
+}
+
+const int8_t* RuntimeContext::getTrimStringOperCharMapById(int id) {
+  return trim_char_maps_->at(id).data();
+}
```
kd97, line 135: need to declare const here?

src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -154,3 +156,97 @@ extern "C" RUNTIME_EXPORT int64_t cider_ascii_upper(int8_t* string_heap_ptr,
 extern "C" void test_to_string(int value) {
   std::printf("test_to_string: %s\n", std::to_string(value).c_str());
 }
+
+extern "C" RUNTIME_EXPORT int64_t cider_concat(char* string_heap_ptr,
+                                               const char* lhs,
+                                               int lhs_len,
+                                               const char* rhs,
+                                               int rhs_len) {
+  StringHeap* ptr = reinterpret_cast<StringHeap*>(string_heap_ptr);
```
kd97, line 165: hint: if it's well to inline the function. better to add key words ALWAYS_INLINE.

kd97 (approved):

src/cider/exec/nextgen/context/CodegenContext.cpp
YBRua: used vector here because we could later pass `int8_t*` pointer to the runtime functions

src/cider/exec/nextgen/context/RuntimeContext.cpp
YBRua: updated

kd97 (dismissed):

jikunshang (approved):

yma11 (approved):

PR #290, opened 2022-12-27
[POAE7-2737] support string type conversion in nextgen

kd97:

cast func in nextgen.

feature migration

No

UTs


kd97:
pls help review @bigPYJ1151 @haiweizh

src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -154,3 +156,139 @@ extern "C" RUNTIME_EXPORT int64_t cider_ascii_upper(int8_t* string_heap_ptr,
...
+  string_t s = ptr->addString(buf, str_len);
+  return pack_string((const int8_t*)s.getDataUnsafe(), (const int32_t)s.getSize());
+}
+
+#define DEF_CONVERT_STRING_TO_INTEGER(value_type, value_name)          \
+  extern "C" RUNTIME_EXPORT value_type convert_string_to_##value_name( \
+      const char* str_ptr, const int32_t str_len) {                    \
+    std::string from_str(str_ptr, str_len);                            \
+    value_type res = std::stoi(from_str);                              \
+    if (res >= std::numeric_limits<value_type>::min() &&               \
```
spevenhe, line 237: The res is ```value_type```, should it always be in ```value_type```'s range?

src/cider/tests/nextgen/functional/CastTest.cpp
```diff
@@ -0,0 +1,143 @@
...
+      QueryArrowDataGenerator::generateBatchByTypes(                                     \
+          schema_,                                                                       \
+          array_,                                                                        \
+          20,                                                                            \
+          {"col_a", "col_b"},                                                            \
+          {CREATE_SUBSTRAIT_TYPE(SUBSTRAIT_TYPE_NAME),                                   \
+           CREATE_SUBSTRAIT_TYPE(SUBSTRAIT_TYPE_NAME)},                                  \
+          {0, 2},                                                                        \
+          GeneratePattern::Random,                                                       \
+          -100,                                                                          \
```
spevenhe, line 40: Do we need to test some failed cases like overflow?

src/cider/exec/nextgen/function/CiderStringFunctions.cpp
spevenhe: maybe need to add a try-catch around stoi

kd97: yes, stoi will return a value of integer type, need value limits check if value type is int16_t/int8_t.

kd97: and just throw runtime exception.

src/cider/tests/nextgen/functional/CastTest.cpp
kd97: you mean primitive type cast check? indeed we need support, I will fix cast overflow check and add related tests after runtime error_code is ready.https://github.com/intel/BDTK/pull/280

src/cider/exec/nextgen/function/CiderStringFunctions.cpp
spevenhe: I think stoi's runtime error would be better wrapped in cider-runtime error

src/cider/type/plan/UnaryExpr.cpp
```diff
@@ -25,50 +25,52 @@ namespace Analyzer {
 using namespace cider::jitlib;
 
 JITValuePointer codegenCastBetweenDateAndTime(CodegenContext& context,
-                                              JITValue& operand_val,
+                                              JITValuePointer operand_val,
                                               const SQLTypeInfo& operand_ti,
                                               const SQLTypeInfo& ti) {
-  JITFunction& func = *context.getJITFunction();
```
bigPYJ1151, line 31: Hmm, why changed this from reference to pointer? You needn't to change anything following if using ```JITFunction& func```.

spevenhe (dismissed):

src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
```diff
@@ -222,7 +222,7 @@ JITValuePointer LLVMJITFunction::emitRuntimeFunctionCall(
 
 void LLVMJITFunction::cloneFunctionRecursive(llvm::Function* fn) {
   CHECK(fn);
-  if (!fn->isDeclaration()) {
+  if (!fn->isDeclaration() || !fn->hasFnAttribute(llvm::Attribute::AlwaysInline)) {
```
yma11, line 225: Is this necessary change?  what will happen if w/o it?

src/cider/type/plan/UnaryExpr.cpp
```diff
@@ -143,60 +174,174 @@ JITExprValue& UOper::codegenUminus(CodegenContext& context, Analyzer::Expr* oper
...
+        "extract_str_ptr",
+        JITFunctionEmitDescriptor{.ret_type = JITTypeTag::POINTER,
+                                  .params_vector = {ptr_and_len.get()}});
+    auto ret_len = func.emitRuntimeFunctionCall(
+        "extract_str_len",
+        JITFunctionEmitDescriptor{.ret_type = JITTypeTag::INT32,
+                                  .params_vector = {ptr_and_len.get()}});
+    return set_expr_value(operand_val.getNull(), ret_len, ret_ptr);
+  } else if (operand_ti.is_string()) {
+    auto str_val = VarSizeJITExprValue(operand->codegen(context));
```
yma11, line 331: need do a check on `target_ti` as it should be numeric?

bigPYJ1151 (approved):

src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
kd97: some external functions that cannot be inlined may core dumped here, so  just clone inline function and use NEVER_INLINE key words in such function declaration to skip it.

src/cider/type/plan/UnaryExpr.cpp
kd97: defined switch-case check in `codegenCastNumericToString`, not supported type will throw exception.

PR #292, opened 2022-12-27
[POAE7-2750] IN expr support for string type in Nextgen

yma11:
 IN expr support for string type in Nextgen

 IN expr support for string type in Nextgen

No

UT

### DIFF
--- a/src/cider/exec/nextgen/context/CiderSet.h
+++ b/src/cider/exec/nextgen/context/CiderSet.h
@@ -167,6 +167,78 @@ class CiderDoubleSet : public CiderSet {
   robin_hood::unordered_set<double> set_;
 };
 
+class CiderStringSet : public CiderSet {
+ public:
+  CiderStringSet() : CiderSet() {}
+
+  void insert(int8_t key_val) {
+    CIDER_THROW(CiderRuntimeException,
+                "CiderStringSet doesn't support insert int8_t value.");
+  }
+
+  void insert(int16_t key_val) {
+    CIDER_THROW(CiderRuntimeException,
+                "CiderStringSet doesn't support insert int16_t value.");
+  }
+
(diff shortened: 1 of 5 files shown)


src/cider/exec/nextgen/context/CiderSet.h
```diff
@@ -167,6 +167,78 @@ class CiderDoubleSet : public CiderSet {
   robin_hood::unordered_set<double> set_;
 };
 
+class CiderStringSet : public CiderSet {
+ public:
+  CiderStringSet() : CiderSet() {}
+
+  void insert(int8_t key_val) {
```
jikunshang, line 174: I think we can impl these `insert` methods in `CiderSet`. No need to make it pure virtual function.

src/cider/exec/nextgen/context/CiderSet.h
```diff
@@ -167,6 +167,78 @@ class CiderDoubleSet : public CiderSet {
...
+
+  bool contains(double key_val) {
+    CIDER_THROW(CiderRuntimeException,
+                "CiderStringSet doesn't support find double value.");
+  }
+
+  bool contains(std::string key_val) { return set_.contains(key_val); }
+
+ private:
+  robin_hood::unordered_set<std::string> set_;
```
jikunshang, line 239: maybe we can use `CiderHashTable` instead in the future?

jikunshang (dismissed):

src/cider/exec/nextgen/context/CiderSet.h
yma11: We may can switch to hashtable if it has better perf. Also we can evaluate more as well when support count distinct.

jikunshang (approved):

PR #298, opened 2022-12-29
[POAE7-2724] replace boost codes by using stl

harborn:


winningsix:
Can you remind me again why we need to replace with stl? You should update PR comments.

jikunshang (approved):

PenghuiJiao (approved):

harborn:
> Can you remind me again why we need to replace with stl? BTW, you should update PR comments.

for upgrading c++20 and gcc10.3 for this project.
currently, bdtk use boost version 1.71
some codes (such as boost::log) will compile fail if we turn on c++20
so, to be more easily maintain the codes, we should use as much as stl api.
and this pr just replacing some boost api to stl api.

winningsix (approved):

PR #299, opened 2022-12-30
[POAE7-2731] JITLib Auto-Vectorize Support: SSE

bigPYJ1151:

JITLib Auto-Vectorize Support: SSE

Add auto-vectorization support.

No

UTs

INFRA

### DIFF
--- a/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITEngine.h
+++ b/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITEngine.h
@@ -29,6 +29,7 @@ class LLVMJITModule;
 // compilation config info
 struct CompilationOptions {
   bool optimize_ir = true;
+  bool enable_vectorize = false;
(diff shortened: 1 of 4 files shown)


src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITEngine.h
```diff
@@ -31,6 +31,7 @@ struct CompilationOptions {
   bool optimize_ir = true;
   bool aggressive_jit_compile = true;
   bool dump_ir = false;
+  bool enable_vectorize = false;
```
winningsix, line 34: Q: why default false?

bigPYJ1151: I think it will increase codegen cost significantly, only enable it when necessary.

Ma-Jian1 (approved):

PR #303, opened 2023-01-03
[POAE7-2767] Fix memory leak in LLVMIfBuilder

bigPYJ1151:

Fix memory leak in LLVMIfBuilder

BugFix

No

UTs

BugFix


winningsix (approved):
Great catch. Need to add some test cases to cover this?

kd97:
bug fixed in my test. q: no need change to `LLVMLoopBuilder`?

kd97 (approved):

bigPYJ1151:
> Great catch. Need to add some test cases to cover this?

Currently address sanitizer is disabled in CI because of some existing memory problems, maybe we should fix them and enable address sanitizer by default.

bigPYJ1151:
> bug fixed in my test. q: no need change to `LLVMLoopBuilder`?

```LoopBuilder``` is fine.

PR #304, opened 2023-01-04
[POAE7-2726] add hashtable builder factory , base interface and example

spevenhe:

add hashtable builder factory , base interface and example

base interface as hashtbale's impl's common api
build factory will be used in hashtable selector in future
add an example to illustrate how hashtable is used

No

Uts

### DIFF
--- a/src/cider/examples/CiderHashTableExample.cpp
+++ b/src/cider/examples/CiderHashTableExample.cpp
@@ -73,7 +73,7 @@ cider_hashtable::HashTableRegistrar<
         void,
         std::allocator<std::pair<cider_hashtable::table_key<int>, std::string>>>,
     cider_hashtable::LinearProbeHashTable<int, std::string, MurmurHash, IntEqual>>
-    linear_hashtable(cider_hashtable::hashtableName::LINEAR_PROBING_INT);
+    linear_hashtable(cider_hashtable::hashtableName::LINEAR_PROBING_STRING);
 
 int main(int argc, char** argv) {
   // query: SELECT t_right.col_2 FROM t_left JOIN t_right ON t_left.col_1 = t_right.col_1;
@@ -107,7 +107,7 @@ int main(int argc, char** argv) {
       void,
       std::allocator<std::pair<cider_hashtable::table_key<int>, std::string>>>>::
                 Instance()
-                    .getHashTable(cider_hashtable::hashtableName::LINEAR_PROBING_INT);
+                    .getHashTable(cider_hashtable::hashtableName::LINEAR_PROBING_STRING);
 
   // build hashtable based on right table, key is col_1, value is col_2
   // For multi project cols in join, the value type can be vector<col1,col2...> or
--- a/src/cider/examples/CiderHashTableExample.cpp
+++ b/src/cider/examples/CiderHashTableExample.cpp
@@ -25,7 +25,6 @@
 #include <random>
 #include <string>
 #include <vector>
-#include "exec/operator/join/CiderLinearProbingHashTable.h"
 #include "exec/operator/join/HashTableFactory.h"
 #include "util/Logger.h"
 
--- a/src/cider/examples/CiderHashTableExample.cpp
+++ b/src/cider/examples/CiderHashTableExample.cpp
@@ -19,7 +19,6 @@
  * under the License.
  */
 
-#include <exec/operator/join/HashTableFactory.h>
 #include <gflags/gflags.h>
 #include <algorithm>
 #include <iostream>
@@ -27,6 +26,7 @@
 #include <string>
 #include <vector>
 #include "exec/operator/join/CiderLinearProbingHashTable.h"
+#include "exec/operator/join/HashTableFactory.h"
 #include "util/Logger.h"
 
 // utils for generate datas
@@ -58,23 +58,6 @@ struct IntEqual {
   bool operator()(int lhs, int rhs) { return lhs == rhs; }
 };
 
-// this part should be done in future hashtable selector's code with strategy, right now
-// is hardcode. BaseHashTable is the base class and LinearProbeHashTable is the impl in
-// this case. key(int) and value(int) data type should be decided by input data type. Hash
-// function(MurmurHash) and equal function(IntEqual) would have default in selector for
-// each data type. Grower and allocator would be default in impl but currently need to
-// copy from impl to set basehashtable
-cider_hashtable::HashTableRegistrar<
-    cider_hashtable::BaseHashTable<
-        int,
-        std::string,
-        MurmurHash,
-        IntEqual,
-        void,
-        std::allocator<std::pair<cider_hashtable::table_key<int>, std::string>>>,
-    cider_hashtable::LinearProbeHashTable<int, std::string, MurmurHash, IntEqual>>
-    linear_hashtable(cider_hashtable::hashtableName::LINEAR_PROBING_STRING);
-
 int main(int argc, char** argv) {
   // query: SELECT t_right.col_2 FROM t_left JOIN t_right ON t_left.col_1 = t_right.col_1;
 
@@ -95,19 +78,22 @@ int main(int argc, char** argv) {
     table_right_col_2[i] = std::to_string(tmp);
   }
 
-  // consturct hashtable from factory.
-  // BaseHashTable is base hashtable interface.
-  // All template arguments for basehashtable would be decided during selector.
-  // int is the key's data type, and since the project is string the value type is string.
-  auto hm = cider_hashtable::HashTableFactory<cider_hashtable::BaseHashTable<
+  // this part should be done in future hashtable selector's code with strategy, right now
+  // is hardcode. BaseHashTable is the base class and LinearProbeHashTable is the impl in
+  // this case. key(int) and value(int) data type should be decided by input data type.
+  // Hash function(MurmurHash) and equal function(IntEqual) would have default in selector
+  // for each data type. Grower and allocator would be default in impl but currently need
+  // to copy from impl to set basehashtable
+  cider_hashtable::HashTableSelector<
       int,
       std::string,
       MurmurHash,
       IntEqual,
       void,
-      std::allocator<std::pair<cider_hashtable::table_key<int>, std::string>>>>::
-                Instance()
-                    .getHashTable(cider_hashtable::hashtableName::LINEAR_PROBING_STRING);
+      std::allocator<std::pair<cider_hashtable::table_key<int>, std::string>>>
+      hashTableSelector;
+  auto hm =
+      hashTableSelector.createForJoin(cider_hashtable::HashTableType::LINEAR_PROBING);
 
   // build hashtable based on right table, key is col_1, value is col_2
   // For multi project cols in join, the value type can be vector<col1,col2...> or
--- a/src/cider/examples/CiderHashTableExample.cpp
+++ b/src/cider/examples/CiderHashTableExample.cpp
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <iostream>
 #include <random>
+#include <string>
 #include <vector>
 #include "exec/operator/join/CiderLinearProbingHashTable.h"
 #include "util/Logger.h"
@@ -66,38 +67,64 @@ struct IntEqual {
 cider_hashtable::HashTableRegistrar<
     cider_hashtable::BaseHashTable<
         int,
-        int,
+        std::string,
         MurmurHash,
         IntEqual,
         void,
-        std::allocator<std::pair<cider_hashtable::table_key<int>, int>>>,
-    cider_hashtable::LinearProbeHashTable<int, int, MurmurHash, IntEqual>>
-    linear_hashtable("linear_int");
+        std::allocator<std::pair<cider_hashtable::table_key<int>, std::string>>>,
+    cider_hashtable::LinearProbeHashTable<int, std::string, MurmurHash, IntEqual>>
+    linear_hashtable(cider_hashtable::hashtableName::LINEAR_PROBING_INT);
 
 int main(int argc, char** argv) {
-  // consturct hashtable from factory
-  // BaseHashTable is base hashtable interface
-  // all template arguments for basehashtable would be decided during selector
+  // query: SELECT t_right.col_2 FROM t_left JOIN t_right ON t_left.col_1 = t_right.col_1;
+
+  // input data
+  // left table has 200 rows; we use it as probe table
+  const int left_table_row_num = 200;
+  std::vector<int> table_left_col_1(left_table_row_num);
+  for (int i = 0; i < left_table_row_num; i++) {
+    table_left_col_1[i] = random(-50, 50);
+  }
+  // right table has 100 rows and two columns; we use it as build table
+  const int right_table_row_num = 100;
+  std::vector<int> table_right_col_1(right_table_row_num);
+  std::vector<std::string> table_right_col_2(right_table_row_num);
+  for (int i = 0; i < right_table_row_num; i++) {
+    int tmp = random(-25, 25);
+    table_right_col_1[i] = tmp;
+    table_right_col_2[i] = std::to_string(tmp);
+  }
+
+  // consturct hashtable from factory.
+  // BaseHashTable is base hashtable interface.
+  // All template arguments for basehashtable would be decided during selector.
+  // int is the key's data type, and since the project is string the value type is string.
   auto hm = cider_hashtable::HashTableFactory<cider_hashtable::BaseHashTable<
       int,
-      int,
+      std::string,
       MurmurHash,
       IntEqual,
       void,
-      std::allocator<std::pair<cider_hashtable::table_key<int>, int>>>>::Instance()
-                .getHashTable("linear_int");
+      std::allocator<std::pair<cider_hashtable::table_key<int>, std::string>>>>::
+                Instance()
+                    .getHashTable(cider_hashtable::hashtableName::LINEAR_PROBING_INT);
 
-  // build hashtable
-  for (int i = 0; i < 100; i++) {
-    int key = random(-25, 25);
-    int value = random(-1000, 1000);
-    hm->emplace(key, value);
+  // build hashtable based on right table, key is col_1, value is col_2
+  // For multi project cols in join, the value type can be vector<col1,col2...> or
+  // tuple<col1, col2..> for row databases. While for column stored databases, the value
+  // can be column's pointer and offset instead of true value. See Uts named
+  // batchAsValueTest in CiderHashTableTest.cpp
+  for (int i = 0; i < right_table_row_num; i++) {
+    hm->emplace(table_right_col_1[i], table_right_col_2[i]);
   }
-
-  // probe hashtable
-  for (int i = 0; i < 100; i++) {
-    int probe_key = random(-100, 100);
+  // do probe
+  for (int i = 0; i < left_table_row_num; i++) {
+    auto probe_key = table_left_col_1[i];
     auto hm_res_vec = hm->findAll(probe_key);
+    // matched one ore more key in build table.
+    // if the value is part of projected columns, it needs to combine with other columns.
+    // if the value is pointer and offset, the true result need further extraction.
+    // in this case, the value is just the project column, so just print it.
     if (hm_res_vec.size() > 0) {
       std::cout << "key:" << probe_key << " value:";
       for (auto value : hm_res_vec) {
--- a/src/cider/examples/CiderHashTableExample.cpp
+++ b/src/cider/examples/CiderHashTableExample.cpp
@@ -51,14 +51,14 @@ struct MurmurHash {
 };
 
 // Different data types should have different equal functions like float and double may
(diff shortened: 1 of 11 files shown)


src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,74 @@
...
+class HashTableRegistrar : public IHashTableRegistrar<HashTableType_t> {
+ public:
+  explicit HashTableRegistrar(std::string name) {
+    HashTableFactory<HashTableType_t>::Instance().registerHashTable(this, name);
+  }
+
+  std::unique_ptr<HashTableType_t> createHashTable();
+};
+}  // namespace cider_hashtable
+#include <exec/operator/join/HashTableFactory.cpp>
```
jikunshang, line 74: why need this 
maybe this is what you need. https://stackoverflow.com/questions/115703/storing-c-template-function-definitions-in-a-cpp-file

src/cider/tests/hashtable/CiderHashTableTest.cpp
```diff
@@ -60,6 +65,43 @@ int random(int low, int high) {
...
+        Equal,
+        void,
+        std::allocator<std::pair<cider_hashtable::table_key<int>,
+                                 std::pair<cider::exec::nextgen::context::Batch*, int>>>>,
+    cider_hashtable::LinearProbeHashTable<
+        int,
+        std::pair<cider::exec::nextgen::context::Batch*, int>,
+        MurmurHash,
+        Equal>>
+    linear_hashtable_batch("linear_batch");
```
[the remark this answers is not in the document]
spevenhe: You mean the name? It will be enum in future, right now we hasn't make it clear about all the options and names. Other hashtables will be added one by one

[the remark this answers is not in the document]
jikunshang: yes. what's your concern to change to enum now?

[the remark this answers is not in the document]
spevenhe: This ```enum``` should only be used by cider itself. For other users, they may not use this enum since the ```value``` and ```allocator``` is only compatible with codegen. I can add one enum which is mainly about hashtable's impl and key data type. Like: linear_hash_int, linear_hash_float, chained_hash_int, ck_int8....

src/cider/exec/operator/join/BaseHashTable.h
```diff
@@ -0,0 +1,65 @@
+/*
```
winningsix, line 1: Move this to discussion regards to package structure.

src/cider/examples/CiderHashTableExample.cpp
```diff
@@ -0,0 +1,136 @@
...
+std::mt19937 gen(rd());
+
+int random(int low, int high) {
+  std::uniform_int_distribution<> dist(low, high);
+  return dist(gen);
+}
+
+// hashfunction would be in a sepearete hashfunction lib in future, right now is hardcode
+// only for int key
+struct MurmurHash {
```
winningsix, line 43: In the end state, do we want to user to specify hash function impl?

src/cider/examples/CiderHashTableExample.cpp
```diff
@@ -0,0 +1,136 @@
...
+    rawHash *= 0xc4ceb9fe1a85ec53L;
+    rawHash ^= unsigned(rawHash) >> 33;
+    return rawHash;
+  }
+};
+
+// Different data types should have different equal functions like float and double may
+// compare only partial precision Different data types. Equal functions would be in a
+// sepearete lib. This is INT equal function.
+struct IntEqual {
```
winningsix, line 57: Do we need an interface for users to impl they own equal methods? Meanwhile, we provide default impl.

spevenhe: Yes. equal function should be a configurable. But  all equal functions I saw are using struct, so there may be some struct example without a virtual class as interface.

src/cider/examples/CiderHashTableExample.cpp
spevenhe: Hash function should always use different default impls for different data types. I think user should not specify the hashfunction impl unless some people invent some new and great hash function that our lib did not contain.

src/cider/exec/operator/join/CiderStdUnorderedHashTable.h
```diff
@@ -18,32 +18,31 @@
...
  public:
   StdMapDuplicateKeyWrapper(){};
-  folly::fbvector<Value> find(Key key) {
+  std::vector<Value> findAll(Key key) {
     auto iter = umap_.find(key);
     if (iter != umap_.end()) {
       return iter->second;
     }
-    folly::fbvector<Value> empty_res;
+    std::vector<Value> empty_res;
```
yma11, line 32: So we switch to use `std::vector` instead of folly one, any background and will there be any performance issue?

src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,83 @@
...
+  const HashTableFactory& operator=(const HashTableFactory&);
+
+  std::map<hashtableName, IHashTableRegistrar<HashTableType_t>*> m_HashTableRegistry;
+};
+
+template <class HashTableType_t, class HashTableImpl_t>
+class HashTableRegistrar : public IHashTableRegistrar<HashTableType_t> {
+ public:
+  explicit HashTableRegistrar(hashtableName name) {
+    HashTableFactory<HashTableType_t>::Instance().registerHashTable(this, name);
```
yma11, line 77: Seems `HashTableRegistrar ` maintain a pair with hash table instance and its name. When this name will be used? any connection with registration in codegen context?

src/cider/exec/operator/join/HashTableFactory.cpp
```diff
@@ -0,0 +1,44 @@
...
+    IHashTableRegistrar<HashTableType_t>* registrar,
+    hashtableName name) {
+  m_HashTableRegistry[name] = registrar;
+}
+
+template <class HashTableType_t>
+std::unique_ptr<HashTableType_t>
+cider_hashtable::HashTableFactory<HashTableType_t>::getHashTable(hashtableName name) {
+  if (m_HashTableRegistry.find(name) != m_HashTableRegistry.end()) {
+    return m_HashTableRegistry[name]->createHashTable();
```
Deegue, line 40: This factory is designed for creating a hash table instead of choosing a hash table to create right? 

In most cases, we only need one hash table instance for join or aggregation. I'm a little bit confused what will be the benifit if we use `registerHashTable` -> `getHashTable` -> `createHashTable` to new a hash table in our scenario.

src/cider/exec/operator/join/BaseHashTable.h
```diff
@@ -0,0 +1,65 @@
...
+  // join Values is vector<Value>)
+  virtual std::vector<Value> findAll(const Key key) = 0;
+  virtual std::vector<Value> findAll(const Key key, size_t hash_value) = 0;
+
+  // For spill cases
+  virtual bool erase(const Key key) = 0;
+  virtual bool erase(const Key key, size_t hash_value) = 0;
+
+  virtual bool contains(const Key key) = 0;
+  virtual bool contains(const Key key, size_t hash_value) = 0;
```
yma11, line 54: Will these APIs be consolidated with the ones used in Agg scenario? Seems APIs are different, underlying layout is different too?

src/cider/exec/operator/join/CiderStdUnorderedHashTable.h
spevenhe: This is based on the feedback from infra team. Folly is a large lib and use anything from folly in codegen context is inappropriate. And std::vector is lighter, so switch to std::vector

src/cider/exec/operator/join/BaseHashTable.h
spevenhe: @Deegue plz take a look. This is the base hashtable interface we discussed in MIP. It should be the common underlying api for join and agg.

src/cider/exec/operator/join/CiderLinearProbingHashTable.h
```diff
@@ -121,15 +123,15 @@ class LinearProbeHashTable {
   using const_iterator = hashtable_iterator<const value_type>;
 
  public:
-  LinearProbeHashTable(size_type bucket_count,
-                       Key empty_key,
+  LinearProbeHashTable(size_type bucket_count = 16,
+                       Key empty_key = NULL,
```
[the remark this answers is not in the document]
spevenhe: For int/float, this null will be zero. This empty's only usage is to get a value of Key's type and use this value to init the hashtable's memory.

src/cider/exec/operator/join/HashTableFactory.cpp
```diff
@@ -0,0 +1,44 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include "cider/CiderException.h"
+
+template <class HashTableType_t, class HashTableImpl_t>
+std::unique_ptr<HashTableType_t>
+cider_hashtable::HashTableRegistrar<HashTableType_t, HashTableImpl_t>::createHashTable() {
+  return std::make_unique<HashTableImpl_t>();
```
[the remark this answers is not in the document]
spevenhe: No. the createhashtable() method would be overload and will add a switch for different hashtable's constructor. I think only for linear probing hashtable, there will be zero arguments

src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+#include <map>
+#include <memory>
+#include <string>
+
+namespace cider_hashtable {
+
+// to be added
+enum hashtableName {
```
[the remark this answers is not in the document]
spevenhe: hmm. in codegen's 
```
  // TBD: HashTable (GroupBy, Join), other objects registration.
  jitlib::JITValuePointer registerBuffer(
      const int32_t capacity,
      const std::string& name = "",
      const BufferInitializer& initializer =
          [](Buffer* buf) { memset(buf->getBuffer(), 0, buf->getCapacity()); },
      bool output_raw_buffer = true);
```
They are calling the ```name```, so I also used ```name```. Type may be confused

src/cider/exec/operator/join/BaseHashTable.h
Deegue: Yes, `AggregatedHashTable` should inherit `BaseHashTable.h` and the memory layout can be different from `JoinHashTable`.

src/cider/exec/operator/join/CiderLinearProbingHashTable.h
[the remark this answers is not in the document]
jikunshang: I'm afraid NULL will not be zero. https://en.cppreference.com/w/cpp/types/NULL

src/cider/exec/operator/join/CiderLinearProbingHashTable.h
```diff
@@ -228,57 +240,27 @@ class LinearProbeHashTable {
...
+  size_t probe_next(size_t idx) const noexcept;
 
  private:
   key_type empty_key_;
   buckets buckets_;
   size_t size_ = 0;
 };
 }  // namespace cider_hashtable
+
+#include <exec/operator/join/CiderLinearProbingHashTable.cpp>
```
qiuyangshen, line 266: Better to use #include "exec/operator/join/CiderLinearProbingHashTable.cpp" for custom files

src/cider/exec/operator/join/BaseHashTable.h
```diff
@@ -0,0 +1,65 @@
...
+  // Improve performance for string cases
+  virtual bool emplace(Key key, Value value, size_t hash_value) = 0;
+
+  // For agg and non-duplicated value cases
+  virtual Value find(const Key key) = 0;
+  virtual Value find(const Key key, size_t hash_value) = 0;
+
+  // For join and duplicated value cases (Values is an internal data structure, for now in
+  // join Values is vector<Value>)
+  virtual std::vector<Value> findAll(const Key key) = 0;
```
chaojun-zhang, line 46: may be use iterator?

src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,83 @@
...
+  std::unique_ptr<HashTableType_t> getHashTable(hashtableName name);
+
+ private:
+  HashTableFactory() {}
+  ~HashTableFactory() { m_HashTableRegistry.clear(); }
+
+  HashTableFactory(const HashTableFactory&);
+  const HashTableFactory& operator=(const HashTableFactory&);
+
+  std::map<hashtableName, IHashTableRegistrar<HashTableType_t>*> m_HashTableRegistry;
```
chaojun-zhang, line 70: use unordered_map

src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,83 @@
...
+  IHashTableRegistrar() {}
+  virtual ~IHashTableRegistrar() {}
+
+ private:
+  IHashTableRegistrar(const IHashTableRegistrar&);
+  const IHashTableRegistrar& operator=(const IHashTableRegistrar&);
+};
+
+template <class HashTableType_t>
+class HashTableFactory {
```
chaojun-zhang, line 51: The purpose of HashTableFactory is used to create a  hashtable intance for either  join or aggregate,  may be we only need two methods , one for join and another for agg just like the velox does.

I can't see any benefit that you use Registrar to registrar a hashTable before use it.

src/cider/exec/operator/join/BaseHashTable.h
spevenhe: I am not quite understand? use iterator as return value?

src/cider/exec/operator/join/HashTableFactory.h
spevenhe: The purpose of HashTableFactory is to hide all hashtable's constructor into this class. The main difference for our hashtable and velox's hashtable is that we have many and they have only one. 
I guess you are thinking making this class looks like this
```
template <typename Key,
          typename Value,
          typename Hash,
          typename KeyEqual,
          typename Grower,
          typename Allocator>
class hashtablefactory 

template <typename... Args>
BaseHashTable<Key, Value, Hash, KeyEqual, Grower, Allocator> createForJoin (HastTableType hashtable_type, Args.. args) {
  switch(hashtable_type)
      case xxx:
          return new xx_hashtable<Key, Value, Hash, KeyEqual, Grower, Allocator>(args);
      case yyy:
          return new yy_hashtable<Key, Value, Hash, KeyEqual, Grower, Allocator>(args);
} 

````

I first write a factory class looked like above. Then I found a way to use a Registrar to registrar so I don't need to add switch branch in the future. 

I don't know which one is better, what is your opinion?

Comment (chaojun-zhang) [reply src/cider/exec/operator/join/HashTableFactory.h:58]: > The purpose of HashTableFactory is to hide all hashtable's constructor into this class. The main difference for our hashtable and velox's hashtable is that we have many and they have only one. I guess you are thinking making this class looks like this
> 
> ```
> template <typename Key,
>           typename Value,
>           typename Hash,
>           typename KeyEqual,
>           typename Grower,
>           typename Allocator>
> class hashtablefactory 
> 
> template <typename... Args>
> BaseHashTable<Key, Value, Hash, KeyEqual, Grower, Allocator> createForJoin (HastTableType hashtable_type, Args.. args) {
>   switch(hashtable_type)
>       case xxx:
>           return new xx_hashtable<Key, Value, Hash, KeyEqual, Grower, Allocator>(args);
>       case yyy:
>           return new yy_hashtable<Key, Value, Hash, KeyEqual, Grower, Allocator>(args);
> } 
> ```
> 
> I first write a factory class looked like above. Then I found a way to use a Registrar to registrar so I don't need to add switch branch in the future.
> 
> I don't know which one is better, what is your opinion?

Inside the 'createForJoin' method , how about to get the hashTable from a pre-registered std::unordered_map<HashTableType,BaseHashTable>?

File: src/cider/exec/operator/join/BaseHashTable.h (hunk 14 above)
Comment (chaojun-zhang) [reply src/cider/exec/operator/join/BaseHashTable.h:49]: > I am not quite understand? use iterator as return value?

Yes

File: src/cider/tests/hashtable/CiderHashTableTest.cpp [hunk 17]
```diff
@@ -68,8 +52,8 @@ TEST(CiderHashTableTest, factoryTest) {
   cider_hashtable::HashTableSelector<
       int,
       int,
-      MurmurHash,
-      Equal,
+      cider_hashtable::MurmurHash,
```
jikunshang, line 55: `using namespace cider_hashtable`  instead?

jikunshang (approved):

src/cider/tests/hashtable/CiderHashTableTest.cpp
spevenhe: In  https://google.github.io/styleguide/cppguide.html#Namespaces says that "Do not use using-directives (e.g., using namespace foo)."

src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,75 @@
...
+  bool operator()(int lhs, int rhs) { return lhs == rhs; }
+};
+
+template <typename Key,
+          typename Value,
+          typename Hash,
+          typename KeyEqual,
+          typename Grower,
+          typename Allocator>
+class HashTableSelector {
```
Deegue, line 57: Also change the filename?

src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,75 @@
...
+ public:
+  // TODO: Define hashtable selection strategy in here
+  HashTableType getHashTableTypeForJoin(std::string data_type,
+                                        int initial_size,
+                                        double ratio) {}
```
Deegue, line 62: A string type `data_type` seems a little bit confusing, maybe it is better to use template for `data_type`?

`initial_size` and `ratio` are optional right? maybe we need to add a default value for them.

Can you add some comments for `ratio`? I have no idea what it stands for and how to set it.

src/cider/exec/operator/join/HashTableFactory.h
```diff
@@ -0,0 +1,75 @@
...
+
+  template <typename... Args>
+  std::unique_ptr<BaseHashTable<Key, Value, Hash, KeyEqual, Grower, Allocator>>
+  createForJoin(HashTableType hashtable_type, Args&&... args);
```
Deegue, line 66: Maybe we should return a HashTable in `getHashTableTypeForJoin` and remove `createForJoin` or rename `createForJoin` to `createHashTable`? If users got `hashtable_type` which means the type of HashTable they should use, they only need a `create` function no matter it is Join or Aggregation.

src/cider/exec/operator/join/HashTableFactory.cpp
```diff
@@ -0,0 +1,44 @@
...
+          typename Allocator>
+template <typename... Args>
+std::unique_ptr<BaseHashTable<Key, Value, Hash, KeyEqual, Grower, Allocator>>
+HashTableSelector<Key, Value, Hash, KeyEqual, Grower, Allocator>::createForJoin(
+    HashTableType hashtable_type,
+    Args&&... args) {
+  switch (hashtable_type) {
+    case LINEAR_PROBING:
+      return std::make_unique<LinearProbeHashTable<Key, Value, Hash, KeyEqual>>(
+          std::forward<Args>(args)...);
```
Deegue, line 38: I'm confused with that `args`. If it stands for arguments in `Strategy2`. we don't need passing it down to HashTable side right? If it's for HashTable use, could you add some comments and examples to explain what may be in `args`?

spevenhe: args are the arguments to hashtable impl's constructor. It can be zero or many

src/cider/exec/operator/join/HashTableFactory.h
spevenhe: besides selector, right now I also put other things like hashfunction in this file. So in future, when we separated  other things  and only leave selector in this class, we can change the filename.

src/cider/exec/operator/join/HashTableFactory.h
spevenhe: This function will be added in future. Right now it is just a placeholder

src/cider/exec/operator/join/HashTableFactory.cpp
Deegue: So it's not the `args` in `Strategy2` for HashTable selection right?
It seems hard to use it since it is template, could you give an example for that?

spevenhe: Ok, I will a example in next pr.

Deegue (approved):
Offline discussed with @spevenhe , code for `JoinHashTable` selector differs from `AggregationHashTable`. The selector for Join will be a separated class out of `JoinHashTable`.

PR #308, opened 2023-01-05
[POAE7-2773] Port ConstantQueryTest to nextgen

yma11:
Port ConstantQueryTest to nextgen and bug fix

Port ConstantQueryTest to nextgen

No

UT


winningsix (dismissed):

src/cider/type/plan/UnaryExpr.cpp
```diff
@@ -162,6 +162,10 @@ JITExprValue& UOper::codegenCast(CodegenContext& context, Analyzer::Expr* operan
   } else {
     // input is primitive type
     FixSizeJITExprValue operand_val(operand->codegen(context));
+    // If arg type is same as target type, erase the cast directly
+    if (get_type_info().get_type() == get_operand()->get_type_info().get_type()) {
+      return set_expr_value(operand_val.getNull(), operand_val.getValue());
+    }
```
kd97, line 168: this is for optimizing or fixing bugs? I think existed decimal cases (simply cast scale field) and time cases (simply cast dimension field) will go down this branch (type is same) and won't do cast.

yma11: This is bug fix for `cast(null as boolean)`. Let me update it.

yma11: The `notnull` property of cast expr is not correct as parsed by substrait-java, so add a step to determine it from its arg instead. The `==` operator API is not supposed to change.
And this check can't be put at head as it doesn't apply on string case, which the `compression` should not be same in some cases. Keep it in primitive cases.

kd97:
```
 if(this->getValueTypeTag() == target_jit_tag){
    return JITValuePointer(this);
  }
```
put it at jit func head can cover this case?
https://github.com/intel/BDTK/blob/ca805e214fd88ed40a8bc37cdfa3c259de7b8d3d/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp#L525-L532

src/cider/type/plan/UnaryExpr.cpp
```diff
@@ -162,6 +162,10 @@ JITExprValue& UOper::codegenCast(CodegenContext& context, Analyzer::Expr* operan
   } else {
     // input is primitive type
     FixSizeJITExprValue operand_val(operand->codegen(context));
```
kd97, line 164: comments is misleading, FixSizeValue included all numeric type, update it in latest pr. https://github.com/intel/BDTK/pull/290

src/cider/type/plan/ConstantExpr.cpp
```diff
@@ -29,7 +29,7 @@ JITExprValue& Constant::codegen(CodegenContext& context) {
     return expr_var;
   }
 
-  auto null = func.createLiteral(JITTypeTag::BOOL, false);
+  auto null = func.createLiteral(JITTypeTag::BOOL, get_is_null());
```
haiweizh, line 32: Is this for generate null constant?

src/cider/type/plan/UnaryExpr.cpp
yma11: I think this judgement should happen in `codegen()` and not be lowered to jitlib. Current change can assure all cases work as expected.

src/cider/type/plan/ConstantExpr.cpp
yma11: yes. In most cases constant is not null but except for `kNull`.

kd97 (approved):

yma11:
@jikunshang @kd97 please help approve again. Thanks.

jikunshang (approved):

kd97 (approved):

PR #311, opened 2023-01-06
[POAE7-2781] primitive type multiple case-when expr support in next codegen

haiweizh:

[POAE7-2618] case/coalesce/if expr support in next codegen.
This story will be divided into four subtasks:

[POAE7-2675] primitive type if-then expr support in next codegen.
[POAE7-2781] primitive type multiple case-when expr support in next codegen.
[POAE7-2782] string type if/case expr support in next codegen.
[POAE7-2783] coalesce expr support in next codegen.
Implement [POAE7-2781] second base on [POAE7-2675].

primitive type multiple case-when expr support in next codegen.

No.

UTs.

cider


src/cider/type/plan/Analyzer.cpp
```diff
@@ -20,9 +20,12 @@
  * under the License.
  */
 #include "type/plan/Analyzer.h"
+#include "exec/nextgen/utils/JITExprValue.h"
 #include "exec/template/DateTimeUtils.h"
+#include "exec/template/IRCodegenUtils.h"
 #include "type/data/sqltypes.h"
```
kd97, line 26: seems no need to include `JITExprValue.h`, also  `IRCodegenUtils.h` is old code path, better to migrate/rename useful func?

src/cider/type/plan/CaseExpr.cpp
```diff
@@ -0,0 +1,277 @@
...
+            *is_case = func.createLiteral(JITTypeTag::BOOL, true);
+          })
+          ->build();
+    }
+    func.createIfBuilder()
+        ->condition([&]() { return !is_case; })
+        ->ifTrue([&]() {
+          cider::exec::nextgen::utils::FixSizeJITExprValue else_jit_expr_value(
+              else_expr->codegen(context));
+          *value = *else_jit_expr_value.getValue();
```
jikunshang, line 254: `value = else_jit_expr_value.getValue();` would be fine? `*` here will bring a memory copy I guess?

src/cider/type/plan/CaseExpr.cpp
```diff
@@ -0,0 +1,277 @@
...
+          *null = *else_jit_expr_value.getNull();
+        })
+        ->build();
+    return set_expr_value(null, value);
+  } else if (case_ti.is_string()) {
+    UNIMPLEMENTED();
```
jikunshang, line 260: add some log or todo

src/cider/type/plan/Analyzer.cpp
haiweizh: Done.

src/cider/type/plan/CaseExpr.cpp
haiweizh: Done.

src/cider/type/plan/CaseExpr.cpp
haiweizh: Yes, this will do a memory copy. Use JITValuePointer.replace instant of * copy.

haiweizh:
This pr will been closed. There is a pr with coalesce, plz visit [POAE7-2783].

PR #312, opened 2023-01-06
[POAE7-2784] Some improvements and clean for code

Deegue:
1. One to one batch checker uses `memcmp` and cannot be ignore order.
2. Allocated memory of `Cells` should be initialized for `isZero()`, which is the same as it is in `FixedHashTable`.
3. Remove some parts we don't need anymore.
4. Fix code conflict of CI for https://github.com/intel/BDTK/pull/317

Improvements.

No.

UTs.


src/cider/tests/utils/CiderBatchChecker.cpp
```diff
@@ -382,7 +382,7 @@ bool CiderBatchChecker::checkArrowEq(
   // This check is only a shortcut
   // even if it fails, it does not necessarily mean the results are wrong
   // we can goto step 4 and check the results row-by-row
-  if (expected_batches.size() == 1 && actual_batches.size() == 1) {
+  if (!ignore_order && expected_batches.size() == 1 && actual_batches.size() == 1) {
```
kd97, line 385: ignore_order will make step3 fail but go through step4? so this is just for skipping useless steps?

src/cider/exec/template/IRCodegen.cpp
```diff
@@ -266,18 +266,16 @@ std::unique_ptr<CodegenColValues> CodeGenerator::codegenConstantExpr(
...
       UNREACHABLE();
     default:
       CHECK_EQ(constant_value.size(), 1);
       return std::make_unique<FixedSizeColValues>(
           constant_value[0],
-          constant_expr->get_is_null()
-              ? llvm::ConstantInt::getTrue(cgen_state_->context_)
-              : llvm::ConstantInt::getFalse(cgen_state_->context_));
+          constant_expr->get_is_null() ? llvm::ConstantInt::getTrue(cgen_state_->context_)
+                                       : nullptr);
```
kd97, line 278: seems old code path, need to change for nextgen?

src/cider/tests/utils/CiderBatchChecker.cpp
Deegue: Yes, it won't affect correctness, but the logic here should be more precise.

src/cider/exec/template/IRCodegen.cpp
Deegue: Yes, am on it, also get some uts failed.

Deegue: Changing this leads to some other errors for handling the nullptr.

Deegue: Reverted this part.

kd97 (dismissed):

YantingTao1315 (approved):

kd97 (approved):

bigPYJ1151 (approved):

PR #314, opened 2023-01-09
[POAE7-2733] JITLib Benchmark: AVX

bigPYJ1151:

JITLib Benchmark: AVX

Performance measure

No

UTs

INFRA


winningsix (approved):

Makefile
```diff
@@ -64,6 +64,7 @@ build-common:
 		  -DENABLE_INTEL_JIT_LISTENER=OFF \
 		  -DPREFER_STATIC_LIBS=OFF \
 		  -DENABLE_ASN=OFF \
+			-DCIDER_ENABLE_AVX512=OFF\
```
xieqi, line 67: use tab instead of space to keep align

bigPYJ1151: I used tab rather than space, seems it is a bug?

xieqi (approved):

PR #342, opened 2023-01-30
[POAE7-2797] jitlib decimal type(int128) support

kd97:

arrow decimal type(int128) & base ops(math/cmp）support in jitlib level

Use llvm Int128 type to impl the underlying calculation of decimal type
complete steps including：
- Aligns the decimal points of lvalue and rvalue by adjusting to the same scale.
- **int128 basic ops**
- Place the right decimal point of the result value.

No

UTs

### DIFF
--- a/cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
+++ b/cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
@@ -122,11 +122,8 @@ llvm::Value* createConstantImpl(llvm::LLVMContext& context, const std::any& valu
   NativeType actual_value = std::any_cast<NativeType>(value);
   if constexpr (std::is_floating_point_v<NativeType>) {
     return getLLVMConstantFP(actual_value, type_tag, context);
-  } else if (JITTypeTraits<type_tag>::bits > 64) {
-    CHECK(JITTypeTraits<type_tag>::bits == 128);
-    return getLLVMConstantInt128(actual_value, type_tag, context);
   } else {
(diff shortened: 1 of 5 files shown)


cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITUtils.h
```diff
@@ -87,6 +89,13 @@ inline llvm::Value* getLLVMConstantInt(uint64_t value,
   }
 }
 
+inline llvm::Value* getLLVMConstantInt128(__int128_t value,
```
bigPYJ1151, line 92: Try to change ```getLLVMConstantInt``` as a template function?

cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITFunction.cpp
```diff
@@ -122,6 +122,9 @@ llvm::Value* createConstantImpl(llvm::LLVMContext& context, const std::any& valu
   NativeType actual_value = std::any_cast<NativeType>(value);
   if constexpr (std::is_floating_point_v<NativeType>) {
     return getLLVMConstantFP(actual_value, type_tag, context);
+  } else if (JITTypeTraits<type_tag>::bits > 64) {
```
winningsix, line 125: Q: why not make condition to ```128 == JITTypeTraits<type_tag>::bits``` directly?

jikunshang (approved):

bigPYJ1151 (approved):

jikunshang (approved):

PR #351, opened 2023-02-02
[POAE7-2817] add chained hashtable

spevenhe:

add chained hashtable implementation

support unknown size conditions

No

UTs

### DIFF
--- a/cpp/src/cider/exec/operator/join/CiderChainedHashTable.cpp
+++ b/cpp/src/cider/exec/operator/join/CiderChainedHashTable.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Intel Corporation.
+ * Copyright(c) 2022-2023 Intel Corporation.
  *
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
@@ -19,17 +19,6 @@
  * under the License.
  */
 
-/*
-A hash map for join. Uses open addressing with linear probing.
-Advantages:
-  - Predictable performance. Doesn'Value use the allocator unless load factor
-    grows beyond 50%. Linear probing ensures cash efficency.
-  - Desgin for no delete/erase action, makes it faster on insert and find
-  - Allow duplicate keys
-Disadvantages:
-  - Significant performance degradation at high load factors.
-  - Maximum load factor hard coded to 50%, memory inefficient.
- */
 #pragma once
 
 namespace cider_hashtable {
--- a/cpp/src/cider/exec/operator/join/CiderChainedHashTable.cpp
+++ b/cpp/src/cider/exec/operator/join/CiderChainedHashTable.cpp
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2022 Intel Corporation.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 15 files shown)


cpp/src/cider/exec/operator/join/CiderChainedHashTable.h
```diff
@@ -0,0 +1,167 @@
...
+  buckets buckets_;
+  size_t size_ = 0;
+};
+}  // namespace cider_hashtable
+
+// separate the implementations into cpp files instead of h file
+// to isolate the implementation from codegen.
+// use include cpp as a method to avoid maintaining too many template
+// declaration in cpp file.
+#include "exec/operator/join/CiderChainedHashTable.cpp"
```
haiweizh, line 167: Delete this and add a line break.

cpp/src/cider/exec/operator/join/CiderChainedHashTable.h
```diff
@@ -0,0 +1,167 @@
...
+
+  bool insert(const std::pair<key_type, Value>& value) {
+    return emplace_impl(value.first, value.second);
+  }
+
+  bool insert(std::pair<key_type, Value>&& value) {
+    return emplace_impl(value.first, std::move(value.second));
+  }
+
+  void emplace(Key key, Value value, bool inserted) {
```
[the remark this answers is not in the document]
spevenhe: updated

cpp/src/cider/exec/operator/join/CiderChainedHashTable.h
spevenhe: updated

cpp/src/cider/exec/operator/join/CiderChainedHashTable.h
```diff
@@ -0,0 +1,167 @@
...
+    while (pow2 < bucket_count) {
+      pow2 <<= 1;
+    }
+    buckets_.resize(pow2);
+  }
+
+  allocator_type get_allocator() const noexcept { return buckets_.get_allocator(); }
+
+  // Capacity
+  bool empty() const noexcept { return size() == 0; }
```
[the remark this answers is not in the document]
spevenhe: Base class is pure virtual class. So I think there is no need.

jikunshang (approved):

cpp/src/cider/exec/processor/DefaultJoinHashTableBuilder.h
```diff
@@ -32,7 +32,12 @@ class DefaultJoinHashTableBuilder : public JoinHashTableBuilder {
  public:
   DefaultJoinHashTableBuilder(const ::substrait::JoinRel& joinRel,
                               const std::shared_ptr<JoinHashTableBuildContext>& context)
-      : joinRel_(joinRel), context_(context) {}
+      : joinRel_(joinRel), context_(context) {
+    // TODO(xinyi): pass some arguments that will decide hashtable type
+    // TODO(xinyi): 1. get the choosed hashtable type
+    // TODO(xinyi): 2. set the hashtable type
+    hashTable_ = std::unique_ptr<JoinHashTable>(new JoinHashTable());
```
qiuyangshen, line 39: It seems that `hashTable_ = std::make_unique<JoinHashTable>();` is a better practice，I can change it in my next PR

cpp/src/cider/exec/operator/join/CiderChainedHashTable.cpp
```diff
@@ -0,0 +1,136 @@
...
+          typename Grower,
+          typename Allocator>
+template <typename K>
+std::vector<Value>
+ChainedHashTable<Key, Value, Hash, KeyEqual, Grower, Allocator>::find_all_impl(
+    const K& key) {
+  std::vector<Value> result;
+  size_t idx = key_to_idx(key);
+  std::vector<value_type> slot = buckets_[idx];
+  for (auto element : slot) {
```
Deegue, line 117: Q: When some values belong to a same key, will it be better if we put all values to a single key slot instead of storing multiple pairs? I mean there will only be one mapping relation of a key in HashTable.

Deegue (approved):

cpp/src/cider/exec/operator/join/CiderChainedHashTable.cpp
spevenhe: This is a trade off between write and read. Making them together is good for probing but needs extra step when build.

PR #352, opened 2023-02-06
[POAE7-2819] Expr Evaluator API

chaojun-zhang:

Implementation of Expression Evaluator API

Add Expression Evaluator API

No

Need add unit test when TODO part is finished.

Cider, publicApi


cpp/src/cider/exec/processor/ExpressionEvaluator.cpp
```diff
@@ -0,0 +1,52 @@
...
+      std::make_shared<BatchProcessorContext>(context->getAllocator());
+  plan::SubstraitPlan substraitPlan{extendedExpression};
+  batchProcessor_ = cider::exec::processor::makeBatchProcessor(substraitPlan.getPlan(),
+                                                               batchProcessorContext);
+}
+
+std::unique_ptr<Batch> ExpressionEvaluator::evaluate(
+    const std::shared_ptr<Batch>& inBatch) {
+  batchProcessor_->processNextBatch(inBatch->getArray(), inBatch->getSchema());
+  batchProcessor_->finish();
```
Ma-Jian1, line 44: Can I call `evaluate` twice?

cpp/src/cider/include/cider/processor/ExpressionEvaluator.h
```diff
@@ -0,0 +1,60 @@
...
+#include "cider/processor/BatchProcessor.h"
+#include "exec/nextgen/context/Batch.h"
+#include "substrait/algebra.pb.h"
+#include "substrait/extended_expression.pb.h"
+
+namespace cider::exec::processor {
+using ExpressionPtr = std::shared_ptr<::substrait::Expression>;
+using ExpressionVector = std::vector<ExpressionPtr>;
+
+using Batch = cider::exec::nextgen::context::Batch;
```
jikunshang, line 35: use `Batch` rather than `ArrowArray/ArrowSchema` as interface?

chaojun-zhang: Use batch would be more convenient than to use pair of ArrowArray/ArrowSchema. we should unify to use Batch instead of array and schema.

cpp/src/cider/exec/processor/ExpressionEvaluator.cpp
chaojun-zhang: invoke 'finish' in destructor method now

cpp/src/cider/include/cider/processor/ExpressionEvaluator.h
Ma-Jian1: users don't want to learn a new type, in the case of having `Arrow` already.

chaojun-zhang: I saw we use the new invent  type ' cider::exec::nextgen::context::Batch',  may be  use the [ArrowBatch ](https://arrow.apache.org/docs/cpp/api/table.html#_CPPv4N5arrow11RecordBatchE)   is make sense.

jikunshang: `RecordBatch` is C++ API. we should not use that.

chaojun-zhang: Any reason we don't want to use the C++ API instead of C ABI?

ci/docker/Dockerfile
```diff
@@ -107,7 +107,7 @@ RUN mkdir -p ~/.gradle/ && \
 # @link: https://github.com/substrait-io/substrait-java/commit/5045aa3840156198d4251d4f3eb79440fc00e0e7
 RUN git clone https://github.com/substrait-io/substrait-java /substrait-java && \
     pushd /substrait-java && \
-    git checkout 5045aa3840156198d4251d4f3eb79440fc00e0e7 && \
+    git checkout 63cfcb33fa7818e91303d79317c12b8489a9014f && \
```
jikunshang, line 110: please also update comments on L106

kyotoyx (approved):

.clang-tidy
```diff
@@ -16,4 +16,4 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-Checks: -*,readability-braces-around-statements,modernize-deprecated-headers,modernize-use-using,modernize-use-emplace,modernize-use-override
+Checks: -*,readability-braces-around-statements,modernize-deprecated-headers,modernize-use-using,modernize-use-emplace,modernize-use-override
```
yma11, line 19: revert this change?

cpp/src/cider/include/cider/processor/ExpressionEvaluator.h
chaojun-zhang: updated

.clang-tidy
chaojun-zhang: updated

ci/docker/Dockerfile
chaojun-zhang: updated

Ma-Jian1 (approved):

PR #354, opened 2023-02-06
[POAE7-2845]The date type in the query is converted to cast as varchar

PenghuiJiao:

The date type in the query is converted to cast as varchar

nightly test failed
velox does not support the date type, and the date type is stored as a varchar type in hadoop, so the date type in the query needs to be converted to the varchar type.
After velox supports the date type, the date-varchar type in hadoop and query needs to be changed back to the date type.

No

cicd

cicd


winningsix:
Is this required by Presto?

jikunshang:
1. what's the root cause of this issue? from your description, `cast as varchar` is your solution, right?
2. any performance drop issue if you change query?

PenghuiJiao:
> 1. what's the root cause of this issue? from your description, `cast as varchar` is your solution, right?
> 2. any performance drop issue if you change query?

velox does not support storing the date type, and now convert the date type to varchar type when storing

PenghuiJiao:
> Is this required by Presto?

Yes, velox does not support date type

harborn (approved):

kyotoyx (approved):

PR #357, opened 2023-02-06
[POAE7-2808] Support Array in ColumnToRowNode&RowToColumnNode

imajiayu:
1. Support **ARRAY** data type in ColumnReader&ColumnWriter.
2. Simply test **ARRAY**.

No

tests/nextgen/compiler/ArrayTest

### DIFF
--- a/cpp/src/cider/exec/module/batch/CiderBatchUtils.cpp
+++ b/cpp/src/cider/exec/module/batch/CiderBatchUtils.cpp
@@ -152,6 +152,8 @@ int64_t getBufferNum(const ArrowSchema* schema) {
         // Struct Type
         case 's':
           return 1;
+        case 'l':
+          return 2;
       }
     case 'u':
       return 3;
@@ -187,6 +189,9 @@ SQLTypes convertArrowTypeToCiderType(const char* format) {
         // Struct Type
         case 's':
           return kSTRUCT;
+        // Array Type
+        case 'l':
+          return kARRAY;
       }
     case 'u':
       return kVARCHAR;
@@ -247,6 +252,8 @@ const char* convertCiderTypeToArrowType(const SQLTypeInfo& sql_info,
           fmt::format("d:{},{}", sql_info.get_precision(), sql_info.get_scale());
       return format_buffer.c_str();
     }
+    case kARRAY:
+      return "+l";
     default:
       CIDER_THROW(CiderCompileException,
                   std::string("Unsupported to convert type ") + toString(type) +
--- a/cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
+++ b/cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
@@ -201,23 +201,24 @@ extern "C" ALWAYS_INLINE int64_t cider_rconcat(char* string_heap_ptr,
 
 extern "C" ALWAYS_INLINE int8_t* get_data_buffer_with_realloc_on_demand(
     const int8_t* input_desc_ptr,
-    const int32_t current_bytes) {
+    const int32_t current_bytes,
+    const int32_t index) {
   const ArrowArray* arrow_array = reinterpret_cast<const ArrowArray*>(input_desc_ptr);
   CiderArrowArrayBufferHolder* holder =
       reinterpret_cast<CiderArrowArrayBufferHolder*>(arrow_array->private_data);
 
   // assumes arrow_array is an array for var-size binary (with 3 buffers)
-  size_t capacity = holder->getBufferSizeAt(2);
+  size_t capacity = holder->getBufferSizeAt(index);
   if (capacity == 0) {
     // initialize buffer with a capacity of 4096 bytes
-    holder->allocBuffer(2, 4096);
+    holder->allocBuffer(index, 4096);
   } else if (current_bytes >= 0.9 * capacity) {
     // double capacity if current bytes take up 90% of capacity
     // assumes we would have enough space for next input after at most one resize op
-    holder->allocBuffer(2, capacity * 2);
+    holder->allocBuffer(index, capacity * 2);
   }
 
-  return holder->getBufferAs<int8_t>(2);
+  return holder->getBufferAs<int8_t>(index);
 }
 
 extern "C" ALWAYS_INLINE int64_t cider_trim(char* string_heap_ptr,
--- a/cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
+++ b/cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
@@ -199,7 +199,7 @@ extern "C" ALWAYS_INLINE int64_t cider_rconcat(char* string_heap_ptr,
   return pack_string_t(s);
 }
 
-extern "C" ALWAYS_INLINE int8_t* get_data_buffer_with_realloc_on_demand(
+extern "C" ALWAYS_INLINE int8_t* get_buffer_with_realloc_on_demand(
     const int8_t* input_desc_ptr,
     const int32_t current_bytes,
     const int32_t index) {
--- a/cpp/src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
+++ b/cpp/src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
@@ -57,6 +57,7 @@ class ColumnReader {
         break;
       case kARRAY:
         readVariableSizeArrayCol(for_null);
+        break;
       default:
         LOG(ERROR) << "Unsupported data type in ColumnReader: "
                    << expr_->get_type_info().get_type_name();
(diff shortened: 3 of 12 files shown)


jikunshang (commented):
possible to add a test use sql as entry?

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -127,6 +132,171 @@ class ColumnWriter {
         setNullBuffer(values.getNull(), for_null), raw_length_buffer, raw_data_buffer);
   }
 
+  int32_t getBytesFromSQLType(SQLTypes type) {
```
imajiayu, line 135: Is there any similar function already exists? If not, where should i move it to?

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -127,6 +132,171 @@ class ColumnWriter {
...
+        "do_memcpy",
+        JITFunctionEmitDescriptor{
+            .ret_type = JITTypeTag::VOID,
+            .params_vector = {
+                raw_data_buffer->castPointerSubType(JITTypeTag::INT8).get(),
+                cur_pointer->castPointerSubType(JITTypeTag::INT8).get(),
+                data_bytes_len->castJITValuePrimitiveType(JITTypeTag::INT32).get()}});
+
+    // set arrow array length only once
+    auto tmp = context_.getJITFunction()->createLocalJITValue([this,
```
imajiayu, line 283: how to do this only once without using `createLocalJITValue`, which must has a return value.

cpp/src/cider/exec/module/batch/CiderBatchUtils.cpp
```diff
@@ -278,6 +280,8 @@ const char* convertCiderTypeToArrowType(SQLTypes type) {
     // timestamp [microseconds]
     case kTIMESTAMP:
       return "tsu";
+    case kARRAY:
+      return "+l";
```
yma11, line 284: need to add case in `convertArrowTypeToCiderType` for `ARRAY` too?

cpp/src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
```diff
@@ -104,6 +107,79 @@ class ColumnReader {
     }
   }
 
+  void readVariableSizeArrayCol(bool for_null) {
+    auto&& [batch, buffers] = context_.getArrowArrayValues(expr_->getLocalIndex());
+    utils::VarSizeArrayExprValue varsize_values(buffers);
+
+    auto& func = batch->getParentJITFunction();
+
+    if (for_null) {
```
yma11, line 116: @Ma-Jian1 for this `for_null` path, is it always the same handle no matter what's the data type is?

cpp/src/cider/type/data/sqltypes.h
```diff
@@ -277,6 +277,10 @@ class SQLTypeInfo {
       , compression(kENCODING_NONE)
       , comp_param(0)
       , size(get_storage_size()) {}
+  SQLTypeInfo(SQLTypes t, SQLTypes st, bool n, const std::vector<SQLTypeInfo>& c)
```
yma11, line 280: I prefer the API should be like `SQLTypeInfo(SQLTypes t, SQLTypeInfo st, bool n` and update the method `get_subtype() ` correspondingly. Current API kind of mix up the Array and Struct definition?

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
yma11: check `inline int64_t getTypeBytes(SQLTypes type) {`?

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -127,6 +132,171 @@ class ColumnWriter {
...
+            if (for_null) {
+              *null_buffer[index_] =
+                  context_.getJITFunction()->createLiteral(JITTypeTag::INT8, 0);
+              return null_buffer;
+            }
+
+            auto len = (arrow_array_len_ + 7) / 3;
+            // memcpy from start to end, do it only once
+            context_.getJITFunction()->emitRuntimeFunctionCall(
+                "do_memcpy",
```
yma11, line 215: same as above.

cpp/src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
yma11: Per offline confirmation, this code path will be removed so we don't need consider for now.

cpp/src/cider/exec/nextgen/operators/ColumnToRowNode.cpp
```diff
@@ -104,6 +107,84 @@ class ColumnReader {
     }
   }
 
+  //| BOOL     | INT64      | INT8*                |   C_TYPE*      |  INT64     |
+  //|----------|------------|----------------------|----------------|------------|
+  //| row_null | row_length | elem_null for column | values for row | row_offset |
```
yma11, line 112: Better to provide an example value about these representations, like using `[[12, -7, 25], null, [0, -127, 127, 50], []]` as in the arrow doc.

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -127,6 +132,178 @@ class ColumnWriter {
...
+        (values.getLength() + values.getOffset()) *
+        utils::getTypeBytes(expr_->get_type_info().getChildAt(0).get_type());
+    auto raw_data_buffer = context_.getJITFunction()->emitRuntimeFunctionCall(
+        "get_data_buffer_with_realloc_on_demand",
+        JITFunctionEmitDescriptor{
+            .ret_type = JITTypeTag::POINTER,
+            .ret_sub_type = JITTypeTag::INT8,
+            .params_vector = {
+                child_array.get(),
+                values_bytes_len->castJITValuePrimitiveType(JITTypeTag::INT32).get(),
```
yma11, line 187: This parameter is given as "current bytes ALREADY wrote in buffer", but here you give "current bytes ALREADY wrote in buffer + current row data length"?

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -127,6 +132,178 @@ class ColumnWriter {
...
+          // set element null buffer
+          // travese all bits for this row
+          auto count =
+              context_.getJITFunction()->createVariable(JITTypeTag::INT32, "count", 0);
+          context_.getJITFunction()
+              ->createLoopBuilder()
+              ->condition([&count, &values]() { return count < values.getLength(); })
+              ->loop([&](LoopBuilder*) {
+                // read
+                auto origin_index = values.getOffset() + count;
```
yma11, line 209: One by one memcpy, right? possible to do memcpy once?

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -201,23 +201,24 @@ extern "C" ALWAYS_INLINE int64_t cider_rconcat(char* string_heap_ptr,
 
 extern "C" ALWAYS_INLINE int8_t* get_data_buffer_with_realloc_on_demand(
```
jikunshang, line 202: maybe we should rename this function?

cpp/src/cider/tests/nextgen/compiler/ArrayTest.cpp
```diff
@@ -0,0 +1,171 @@
...
+  generator::SubstraitToRelAlgExecutionUnit substrait2eu(plan);
+  auto eu = substrait2eu.createRelAlgExecutionUnit();
+
+  // Pipeline Building
+  auto pipeline = parsers::toOpPipeline(eu);
+  transformer::Transformer transformer;
+  return transformer.toTranslator(pipeline);
+}
+
+context::RuntimeCtxPtr executeAndReturnRuntimeCtx(const std::string& create_ddl,
```
jikunshang, line 59: why not return outBatch directly?

cpp/src/cider/tests/nextgen/compiler/ArrayTest.cpp
```diff
@@ -0,0 +1,171 @@
...
+  // Execution
+  auto runtime_ctx = codegen_ctx.generateRuntimeCTX(allocator);
+  query_func((int8_t*)runtime_ctx.get(), (int8_t*)array);
+
+  return runtime_ctx;
+}
+
+class CiderArrayTest : public ::testing::Test {
+ public:
+  template <typename COL_TYPE>
```
jikunshang, line 97: need to be template function? or we should use template type rather than int8_t in function body?

cpp/src/cider/tests/nextgen/compiler/ArrayTest.cpp
imajiayu: outBatch will be released when runtimeContext is released.

yma11 (approved):

jikunshang (approved):

PR #359, opened 2023-02-07
[POAE7-2848] Fix bugs in cast bool expr

kd97:

- support cast bool type. In some cases it will be converted to case...when... expr, use generated json file to test direct bool conversion.
- clean up code

pass presto e-2-e sql tests. @PenghuiJiao  pls take a review.

No

UTs

### DIFF
--- a/cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
+++ b/cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
@@ -541,8 +541,9 @@ JITValuePointer LLVMJITValue::castJITValuePrimitiveType(JITTypeTag target_jit_ta
   llvm::Value* source_lv = load();
   llvm::Value* target_lv = nullptr;
   if (source_type->isIntegerTy() && target_type->isIntegerTy()) {
-    target_lv =
-        getFunctionBuilder(parent_function_).CreateIntCast(source_lv, target_type, true);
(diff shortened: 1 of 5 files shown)


cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITValue.cpp
```diff
@@ -541,8 +541,9 @@ JITValuePointer LLVMJITValue::castJITValuePrimitiveType(JITTypeTag target_jit_ta
   llvm::Value* source_lv = load();
   llvm::Value* target_lv = nullptr;
   if (source_type->isIntegerTy() && target_type->isIntegerTy()) {
-    target_lv =
-        getFunctionBuilder(parent_function_).CreateIntCast(source_lv, target_type, true);
+    target_lv = getFunctionBuilder(parent_function_)
+                    .CreateIntCast(
+                        source_lv, target_type, source_type->getScalarSizeInBits() != 1);
```
Deegue, line 546: Q: Here we should get whether `source_lv` is signed or not, right? If so, does `source_type->getScalarSizeInBits() != 1` mean that?

kd97: see difference between 'zext .. to' and 'sext .. to' 
https://releases.llvm.org/2.6/docs/LangRef.html#convertops

PenghuiJiao (approved):

jikunshang (approved):

PR #360, opened 2023-02-07
[POAE7-2841] Add UInt128 and UInt256 support for HashTable and Agregation

Deegue:
1. Use `wide_integer.h` of Github project [wide_integer](https://github.com/cerevra/int), which can be used in any bit size.
2. Wrap the implementation as: 
`integer<128, signed>` as `Int128`, 
`integer<128, unsigned>` as `UInt128`, 
`integer<256, signed>` as `Int256`, 
`integer<256, unsigned>` as `UInt256`.
3. Add related HashTable implementation and UTs.

Support key types Int128 and Int256.

No.

UTs.


cpp/src/cider/type/data/sqltypes.h
```diff
@@ -70,6 +70,8 @@ enum SQLTypes {
   kCOLUMN = 27,
   kCOLUMN_LIST = 28,
   kSTRUCT = 29,
+  kINT128 = 30,
+  kINT256 = 31,
   kSQLTYPE_LAST
```
kd97, line 75: internal type? I think it's not sql type.

Deegue: Well, it's about the naming, right? Maybe we can call large int or huge int etc.

Deegue: `kINT128` and `kINT256` will be used as key of aggregation and they're new column types.

kd97: I think it belongs to type implementation. better use get_logical_size() to identify internal type？

kd97: need to update sqltypeInfo in sqltypes.h?

Deegue: I guess not in this PR, since we can add this while supporting Int128&Int256 for the whole workflow.

cpp/src/cider/common/hashtable/Hash.h
```diff
@@ -237,15 +227,96 @@ DEFINE_HASH(uint8_t)
...
+struct UInt128HashCRC32 {
+  size_t operator()(UInt128 x) const {
+    UInt64 crc = -1ULL;
+    crc = _mm_crc32_u64(crc, x.items[0]);
+    crc = _mm_crc32_u64(crc, x.items[1]);
+    return crc;
+  }
+};
+
+#elif defined(__aarch64__) && defined(__ARM_FEATURE_CRC32)
```
jikunshang, line 260: I think we can remove this.

cpp/src/cider/common/interpreters/AggregationHashTable.h
```diff
@@ -190,6 +198,8 @@ class AggregationHashTable final {
   AggregatedHashTableWithUInt16Key agg_ht_uint16_;
   AggregatedHashTableWithUInt32Key agg_ht_uint32_;
   AggregatedHashTableWithUInt64Key agg_ht_uint64_;
+  AggregatedHashTableWithKeys128 agg_ht_uint128_;
```
jikunshang, line 201: it's Uint128 hashmap or int128 hashmap?

cpp/src/cider/common/base/wide_integer.h
```diff
@@ -0,0 +1,352 @@
...
+
+template <typename Arithmetic, size_t Bits, typename Signed>
+struct common_type<Arithmetic, cider::wide::integer<Bits, Signed>>;
+
+}  // namespace std
+
+namespace cider::wide {
+
+template <size_t Bits, typename Signed>
+class integer {
```
jikunshang, line 75: integer -> Integer

cpp/src/cider/common/hashtable/Hash.h
Deegue: Removed.

cpp/src/cider/common/interpreters/AggregationHashTable.h
Deegue: It's UInt128 and UInt256, PR&Jira titles changed.

Deegue: But it can be used for short strings, so it's named `Keys128`.

jikunshang: And I think we should not name these classes as  `AggregatedHashTableWithXXXkey`, `AggregatedHashTableForXXXkey` would be better.

Deegue: Emm, do we need to change all names above from `AggregatedHashTableWithXXX` to `AggregatedHashTableForXXX` ?

jikunshang: I prefer to do so. you can pull someone else to give more feedback about naming.

Deegue: I also prefer `for` rather than `with`, renamed.

cpp/src/cider/common/base/wide_integer.h
Deegue: Done.

cpp/src/cider/common/hashtable/Hash.h
```diff
@@ -237,15 +219,72 @@ DEFINE_HASH(uint8_t)
...
 
 #undef DEFINE_HASH
 
+struct UInt128Hash {
+  size_t operator()(UInt128 x) const {
+    return CityHash_v1_0_2::Hash128to64({x.items[0], x.items[1]});
+  }
+};
+
+#ifdef __SSE4_2__
```
harborn, line 241: 这些code是手写的还是抄的？
这个代码库的代码已经快失控了，能少抄就少抄吧

Deegue: Well, you can find some places using `__SSE4_2__` which are brought from ck.
I think we will get better performance by using `SIMD` or other instructions for accelerating.
BTW, it's not the first place `__SSE4_2__` is used in our project, we can also find it in third-party libs like `velox` and `rapidjson`.

kd97 (approved):

jikunshang (approved):

PR #365, opened 2023-02-09
[POAE7-2857]destroy stringheap after processNextBatch call 

jikunshang:

destroy stringheap after processNextBatch call to reduce memory consumption.

reduce memory consumption for string functions.

No

ut

### DIFF
--- a/cpp/src/cider/exec/processor/DefaultBatchProcessor.cpp
+++ b/cpp/src/cider/exec/processor/DefaultBatchProcessor.cpp
@@ -91,7 +91,7 @@ void DefaultBatchProcessor::processNextBatch(const struct ArrowArray* array,
     CIDER_THROW(CiderRuntimeException,
                 getErrorMessageFromErrCode(static_cast<cider::jitlib::ERROR_CODE>(ret)));
   }
-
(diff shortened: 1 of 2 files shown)


Deegue (approved):

cpp/src/cider/exec/processor/DefaultBatchProcessor.cpp
```diff
@@ -91,7 +91,7 @@ void DefaultBatchProcessor::processNextBatch(const struct ArrowArray* array,
     CIDER_THROW(CiderRuntimeException,
                 getErrorMessageFromErrCode(static_cast<cider::jitlib::ERROR_CODE>(ret)));
   }
-
+  runtime_context_->destroyStringHeap();
```
chaojun-zhang, line 94: may be expose a more generic method named `destroy` and  invoke `destroyStringHeap` inside runtimeContext?

jikunshang: `destroyStringHeap` will be called for each `processNextBatch` call. while `destroy` should be called after the last `processNextBatch` call.

yma11 (approved):

PR #384, opened 2023-02-21
[POAE7-2877]PipelineOperator performance test against velox

Ma-Jian1:
PipelineOperator performance test against velox
 and try to remove schema import which boost 7%

```bash
============================================================================
/workspace/code/BDTK/cpp/src/cider-velox/benchmark/expression/PipelineOperator.cpprelative  time/iter  iters/s
============================================================================
velox                                                        15.63s   63.96m
nextgen                                          130.69%     11.96s   83.59m
nextgenOpt                                       137.45%     11.38s   87.91m
```

performance test

No

no need

INFRA

### DIFF
--- a/cpp/src/cider-velox/benchmark/expression/PipelineOperator.cpp
+++ b/cpp/src/cider-velox/benchmark/expression/PipelineOperator.cpp
@@ -79,7 +79,7 @@ inline CodegenOptions getBaseOption() {
 
 class PipelineOperator : public functions::test::FunctionBenchmarkBase {
  public:
-  explicit PipelineOperator() : FunctionBenchmarkBase() {
+  PipelineOperator() : FunctionBenchmarkBase() {
     // registerAllScalarFunctions() just register checked version for integer
     // we register uncheck version for integer and checkedPlus for compare
     registerFunction<MultiplyFunction, int8_t, int8_t, int8_t>({"multiply"});
--- a/cpp/src/cider-velox/benchmark/expression/PipelineOperator.cpp
+++ b/cpp/src/cider-velox/benchmark/expression/PipelineOperator.cpp
@@ -19,13 +19,9 @@
  * under the License.
  */
 
-#include <memory>
-#include "velox/core/PlanNode.h"
-#include "velox/exec/FilterProject.h"
-#include "velox/vector/ComplexVector.h"
-#define SHARED_LOGGER_H  // ignore util/Logger.h
 #include <folly/Benchmark.h>
 #include <gflags/gflags.h>
+#include <memory>
 
 #include "Allocator.h"
 #include "cider/CiderOptions.h"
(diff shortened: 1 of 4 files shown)


cpp/src/cider-velox/src/CiderPipelineOperator.cpp
```diff
@@ -43,10 +43,8 @@ void CiderPipelineOperator::addInput(RowVectorPtr input) {
   this->input_ = std::move(input);
   ArrowArray* inputArrowArray = CiderBatchUtils::allocateArrowArray();
   exportToArrow(input_, *inputArrowArray);
-  ArrowSchema* inputArrowSchema = CiderBatchUtils::allocateArrowSchema();
-  exportToArrow(input_, *inputArrowSchema);
 
-  batchProcessor_->processNextBatch(inputArrowArray, inputArrowSchema);
```
chaojun-zhang, line 49: talked with xieqi before, we may need the schema to do validate staff.

cpp/src/cider-velox/benchmark/expression/PipelineOperator.cpp
```diff
@@ -0,0 +1,265 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <folly/Benchmark.h>
+#include <gflags/gflags.h>
+#include <memory>
+
+#include "Allocator.h"
```
xieqi, line 26: include full path

cpp/src/cider-velox/src/CiderPipelineOperator.cpp
xieqi: since your benchmark doesn't use this file, maybe a comment is better, change the public API need a discussion through we don't use the schema in the cider now.

xieqi: since your benchmark doesn't use this file a comment is better through cider doesn't use schema now.

Ma-Jian1: since `try to remove schema import which boost 7%`, validating the schema for every batch seems to have no benefit, and velox also doesn't validate every row vector.

yma11: seems the encoded string support need the schema info, can it be exported once and kept/passed to cider?

Ma-Jian1: @jikunshang seems does not check schema？ just check `dictionary` in Array. 
If we need schema once, should split it from `processNextBatch`.

jikunshang: yes, you are right. I remember we will not pass `ArrowSchema` to runtime context.

yma11: Then it makes sense to hack the schema.

yma11 (approved):

chaojun-zhang (approved):

PR #386, opened 2023-02-23
[POAE7-2881]enable auto-vectorization for UOper，Constant

Ma-Jian1:
enable auto-vectorization for UOper, Constant, and divide

support auto-vectorization

No

benchmark test

INFRA


cpp/src/cider/type/plan/BinaryExpr.cpp
```diff
@@ -36,6 +36,7 @@ void BinOper::initAutoVectorizeFlag() {
       case kPLUS:
       case kMINUS:
       case kMULTIPLY:
+      case kDIVIDE:
```
bigPYJ1151, line 39: ```Div``` can't be vectorizable.

Ma-Jian1: enable Div for complex arithmetic operations, `i8*i16*2 + i16/3 - 1` gain 600%+ boost.
if we don't enable Div，`i8*i16*2 + i16/3 - 1` is slow than velox.

bigPYJ1151: Hm, I found div with literal is vectorizable, interesting.

Ma-Jian1: ok, I'll narrow auto-vectorization if the dividend is constant.

bigPYJ1151 (approved):

winningsix (approved):

PR #387, opened 2023-02-27
[POAE7-2888] nextgen string function refactor

jikunshang:

refactor string function in nextgen codege. move memory allocation, memory copy out of main for loop.

for better performance.

No

UT.

### DIFF
--- a/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
+++ b/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
@@ -83,7 +83,7 @@ class StringFunctionBenchmark : public functions::test::FunctionBenchmarkBase {
   explicit StringFunctionBenchmark(size_t vectorSize) : FunctionBenchmarkBase() {
     functions::prestosql::registerStringFunctions();
 
-    inputType_ = ROW({{"col_str", VARCHAR()}});
+    inputType_ = ROW({{"col_str", VARCHAR()},{"col_str2", VARCHAR()}});
 
     vectorSize_ = vectorSize;
     // Generate input data.
@@ -93,6 +93,7 @@ class StringFunctionBenchmark : public functions::test::FunctionBenchmarkBase {
     VectorFuzzer fuzzer(opts, pool(), FLAGS_fuzzer_seed);
 
     std::vector<VectorPtr> children;
+    children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));  // varchar
     children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));  // varchar
 
     rowVector_ = std::make_shared<RowVector>(
@@ -194,7 +195,8 @@ std::unique_ptr<StringFunctionBenchmark> benchmark;
 // }
 // BENCHMARK_DRAW_LINE();
 
-std::string concat_expr = "concat(col_str, 'aaaa')";
+// std::string concat_expr = "concat(col_str, 'aaaa')";
+std::string concat_expr = "concat(col_str, col_str2)";
 
 BENCHMARK(nextgen_concat) {
   benchmark->nextgenCompute(concat_expr);
--- a/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
+++ b/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
@@ -19,7 +19,6 @@
  * under the License.
  */
 
-#define SHARED_LOGGER_H  // ignore util/Logger.h
 #include <folly/Benchmark.h>
 #include <gflags/gflags.h>
 
--- a/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
+++ b/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
@@ -61,8 +61,8 @@ using namespace facebook::velox::functions;
 using namespace facebook::velox::exec::test;
 using namespace facebook::velox::substrait;
 
-static const std::shared_ptr<CiderAllocator> allocator =
-    std::make_shared<CiderDefaultAllocator>();
+// static const std::shared_ptr<PoolAllocator> allocator =
+//     std::make_shared<CiderDefaultAllocator>();
 namespace {
 std::pair<ArrowArray*, ArrowSchema*> veloxVectorToArrow(RowVectorPtr vec,
                                                         MemoryPool* pool) {
@@ -83,13 +83,15 @@ class StringFunctionBenchmark : public functions::test::FunctionBenchmarkBase {
   explicit StringFunctionBenchmark(size_t vectorSize) : FunctionBenchmarkBase() {
     functions::prestosql::registerStringFunctions();
 
-    inputType_ = ROW({{"col_str", VARCHAR()},{"col_str2", VARCHAR()}});
+    inputType_ = ROW({{"col_str", VARCHAR()}, {"col_str2", VARCHAR()}});
 
     vectorSize_ = vectorSize;
     // Generate input data.
     VectorFuzzer::Options opts;
     opts.vectorSize = vectorSize;
-    opts.nullRatio = 0.5;
+    // opts.nullRatio = 0.5;
+    // opts.stringVariableLength = false;
+    // opts.stringLength = 16;
     VectorFuzzer fuzzer(opts, pool(), FLAGS_fuzzer_seed);
 
     std::vector<VectorPtr> children;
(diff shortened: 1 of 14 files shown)


cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -199,7 +248,69 @@ extern "C" ALWAYS_INLINE int64_t cider_rconcat(char* string_heap_ptr,
...
+extern "C" ALWAYS_INLINE void copy_string_buffer(const int8_t* input_desc_ptr,
+                                                 int8_t* buffer,
+                                                 int64_t total_row) {
+  const int64_t* offset_buffer = reinterpret_cast<const int64_t*>(
+      reinterpret_cast<const ArrowArray*>(input_desc_ptr)->buffers[1]);
+  int32_t* actual_writable_offset_buffer =
+      reinterpret_cast<int32_t*>(const_cast<int64_t*>(offset_buffer));
+  int32_t cur_offset = 0;
+  for (int i = 0; i < total_row; i++) {
+    int32_t cur_len = (offset_buffer[i] >> 48);
```
Ma-Jian1, line 282: encode length and offset into one int64_t? maybe you should define some util macro explicitly.
and maybe 32 for length and offset be better?

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -211,11 +322,11 @@ extern "C" ALWAYS_INLINE int8_t* get_buffer_with_realloc_on_demand(
   size_t capacity = holder->getBufferSizeAt(index);
   if (capacity == 0) {
     // initialize buffer with a capacity of 4096 bytes
```
Ma-Jian1, line 324: fix comment.

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -419,7 +461,41 @@ void RowToColumnTranslator::codegenImpl(SuccessorEmitter successor_wrapper,
       auto&& [arrow_array, _] = context.getArrowArrayValues(local_offset);
       codegen_utils::setArrowArrayLength(arrow_array, output_index);
     }
-  });
+    // for string expressions, we dump the final buffer here.
+    for (auto& expr : output_exprs) {
+      if (auto strExpr = std::dynamic_pointer_cast<Analyzer::StringOper>(expr)) {
+        if (!strExpr->isOutput()) {
+          break;
```
Ma-Jian1, line 468: continue? or do you sort output_exprs?

cpp/src/cider/type/plan/StringOpExpr.h
```diff
@@ -367,6 +370,7 @@ class CharLengthStringOper : public StringOper {
       : StringOper(string_oper) {}
 
   std::shared_ptr<Analyzer::Expr> deep_copy() const override;
+  void setIsOutput() override { is_output_ = false; }
```
Ma-Jian1, line 373: Could you add an argument? It's weird to set it to false.

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
jikunshang: it's string pointer and length.

cpp/src/cider/type/plan/StringOpExpr.h
jikunshang: emmm, `is_output_` means whether this expression is the output expression and should this expression do string type output write. for example `concat(substr(col_1, 1, 2), 'abc')`, `concat` should be a output expression and it should do string type output write. While for `charLength(concat(col_1, col2))` , `charLength` is a output expression but it's output is a integer rather than a string, so it should not do extra string write after for loop.

actually, we should have 2 flag. one is is_output_, this should be set via plan side. the other flag is whether return string.

cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
```diff
@@ -0,0 +1,221 @@
...
+//     std::make_shared<CiderDefaultAllocator>();
+namespace {
+std::pair<ArrowArray*, ArrowSchema*> veloxVectorToArrow(RowVectorPtr vec,
+                                                        MemoryPool* pool) {
+  for (size_t i = 0; i < vec->childrenSize(); i++) {
+    vec->childAt(i)->mutableRawNulls();
+  }
+  ArrowArray* inputArrowArray = CiderBatchUtils::allocateArrowArray();
+  exportToArrow(vec, *inputArrowArray, pool);
+  ArrowSchema* inputArrowSchema = CiderBatchUtils::allocateArrowSchema();
```
yma11, line 73: Should the schema convertor also can be removed like in expression benchmark?

cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
```diff
@@ -0,0 +1,221 @@
...
+    children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));  // varchar
+
+    rowVector_ = std::make_shared<RowVector>(
+        pool(), inputType_, nullptr, vectorSize, std::move(children));
+
+    ArrowSchema* _schema;
+    std::tie(inputArray_, _schema) = veloxVectorToArrow(rowVector_, execCtx_.pool());
+    // hack: make processor don't release inputArray_, otherwise we can't use inputArray_
+    // multi times.
+    inputReleaser_ = inputArray_->release;
```
yma11, line 107: So we use same batch multi times? will it affect benchmark result?

cpp/src/cider/exec/nextgen/operators/RowToColumnNode.cpp
```diff
@@ -96,39 +101,76 @@ class ColumnWriter {
...
-            .ret_type = JITTypeTag::VOID,
-            .params_vector = {
-                cur_pointer.get(), values.getValue().get(), values.getLength().get()}});
+  void writeVariableSizeTypeColForStrExpr() {
+    // Get values need to write
+    utils::FixSizeJITExprValue values(expr_->get_expr_value());
+
+    // we will borrow offset buffer to store intermiedate result(string ptr and
+    // length, 64 bits), so allocate (arrow_array_len_ * 8) bytes, actual offset size
+    // should be (arrow_array_len_ + 1) * 4 bytes
```
yma11, line 160: Seems it's an intermediate and temporary buffer to store the string ptr and length, instead of previous `raw_length_buffer` and `raw_data_buffer`. I think you don't need above comments and rename `actual_raw_length_buffer` to `raw_str_ptr_buffer`?

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
Ma-Jian1: If it's an actual pointer， then encoding the pointer into 48bit is risky.

cpp/src/cider/type/plan/StringOpExpr.h
Ma-Jian1: I mean `void setIsOutput(bool val)`

jikunshang: you mean 
`void StringOper::setIsOutput(bool is_output) { is_output_ = is_output; }`
`void CharLengthStringOper::setIsOutput(bool is_output) { is_output_ = false; }`

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
jikunshang: The [x86-64 architecture](https://en.wikipedia.org/wiki/X86-64) (as of 2016) allows 48 bits for virtual memory and, for any given processor, up to 52 bits for physical memory.[[27]](https://en.wikipedia.org/wiki/64-bit_computing#cite_note-28)[[28]](https://en.wikipedia.org/wiki/64-bit_computing#cite_note-29). I think there is no risk for recent CPUs.

cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
jikunshang: yes. I think it won't.

cpp/src/cider/type/plan/StringOpExpr.h
Ma-Jian1: `CharLengthStringOper` inherit from `StringOper`, which doesn't need to overwrite the `setIsOutput()` definition, just call `setIsOutput(false)`

Ma-Jian1 (dismissed):

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -177,6 +207,25 @@ extern "C" ALWAYS_INLINE int64_t cider_concat(char* string_heap_ptr,
   return pack_string_t(s);
 }
 
+extern "C" ALWAYS_INLINE int32_t cider_concat_len(int lhs_len, int rhs_len) {
```
yma11, line 210: What's the purpose for functions `cider_concat_len`, `cider_ascii_upper_len` and `cider_ascii_lower_ptr`? seems they are not necessary.

jikunshang: yes, it's not necessary, we can calculate the result size directly in xxxStringExpr::codegen method. keep these functions may help understand. By the way, this functions can be optimized.

Ma-Jian1 (approved):

yma11 (approved):

PR #390, opened 2023-02-27
[POAE7-2890] Fix core dump and improvement of HashMap

Deegue:
1. Fix core dump when doing `resize()`.
2. Set default size of HashMap to 1 << 12.
3. Optimize HashTableBenchmark with larger number of input rows, and get the performance result:

image

Fix & improve.

No.

UTs.

### DIFF
--- a/cpp/src/cider/benchmarks/HashMapBenchmark.cpp
+++ b/cpp/src/cider/benchmarks/HashMapBenchmark.cpp
@@ -151,5 +151,4 @@ BENCHMARK(BM_Optimized_Lookup<uint8_t>)->RangeMultiplier(10)->Range(100, 10000);
 BENCHMARK(BM_Baseline_Lookup<uint16_t>)->RangeMultiplier(10)->Range(100, 10000);
 BENCHMARK(BM_Optimized_Lookup<uint16_t>)->RangeMultiplier(10)->Range(100, 10000);
 
-
 BENCHMARK_MAIN();
--- a/cpp/src/cider/benchmarks/HashMapBenchmark.cpp
+++ b/cpp/src/cider/benchmarks/HashMapBenchmark.cpp
@@ -113,11 +113,11 @@ static std::tuple<std::vector<T>, std::vector<bool>> generateAndFillVector(
 
 template <typename KeyType, typename HashTableType>
 static void BM_Lookup(benchmark::State& state) {
-  size_t row_num = 100;
+  size_t row_num = state.range(0);
   KeyType value_min = std::numeric_limits<KeyType>::min();
   KeyType value_max = std::numeric_limits<KeyType>::max();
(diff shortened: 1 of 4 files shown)


cpp/src/cider/benchmarks/HashMapBenchmark.cpp
```diff
@@ -113,11 +113,11 @@ static std::tuple<std::vector<T>, std::vector<bool>> generateAndFillVector(
 
 template <typename KeyType, typename HashTableType>
 static void BM_Lookup(benchmark::State& state) {
-  size_t row_num = 100;
+  size_t row_num = state.range(0);
   KeyType value_min = std::numeric_limits<KeyType>::min();
   KeyType value_max = std::numeric_limits<KeyType>::max();
   size_t null_chance = 0;
-  GeneratePattern pattern = GeneratePattern::Random;
+  GeneratePattern pattern = GeneratePattern::Sequence;
```
jikunshang, line 120: why change to sequence?

jikunshang:
why it will core dump when doing resize?

cpp/src/cider/benchmarks/HashMapBenchmark.cpp
Deegue: We can find `collision` which is almost a fixed value, to better illustrate the performance of hashtables.

Deegue:
> why it will core dump when doing resize?

Core dump occurred on `emplace` while `resize` is not correct.

jikunshang:
> > why it will core dump when doing resize?
> 
> Core dump occurred on `emplace` while `resize` is not correct.

So the root cause is `resize` have bugs? If so, better add some ut for resize? Please point out the root cause,and how you fix it in PR description.

Deegue:
> > > why it will core dump when doing resize?
> > 
> > 
> > Core dump occurred on `emplace` while `resize` is not correct.
> 
> So the root cause is `resize` have bugs? If so, better add some ut for resize? Please point out the root cause,and how you fix it in PR description.

Added a larger dataset test case.

jikunshang (approved):

harborn (approved):

cpp/src/cider/common/hashtable/HashTable.h
```diff
@@ -251,7 +251,7 @@ void insertSetMapped(MappedType& dest, const ValueType& src) {
  * Has very small state (one uint8_t) and useful for Set-s allocated in automatic memory
  * (see uniqExact as an example).
  */
-template <size_t initial_size_degree = 8>
+template <size_t initial_size_degree = 12>
```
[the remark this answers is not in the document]
Deegue: Not yet, I think it's a good idea and maybe we can add memory monitor in `CiderAllocator`.

PR #393, opened 2023-02-28
[POAE7-2894]recover cider plan validator and scalar function support

yma11:
1. recover plan validator module and UT by isolating function look up as a dedicate module
2. fix external function support(between and)
3. remove some string expr unused code.

UT fix

No

UT

coderefactor


harborn (dismissed):

cpp/src/cider/function/scalar/ExtensionFunctions.hpp
```diff
@@ -25,8 +25,8 @@
 #include <cmath>
 #include <cstdint>
 #include <cstdlib>
-#include "type/data/CiderTypes.h"
-#include "type/data/funcannotations.h"
+#include "../../type/data/CiderTypes.h"
```
jikunshang, line 28: absolute path instead?

cpp/src/cider/function/scripts/parse_ast.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import print_function
```
jikunshang, line 1: when will this be used?

cpp/src/cider/type/data/CiderTypes.h
```diff
@@ -0,0 +1,40 @@
+/*
```
jikunshang, line 1: rename this file?

ci/scripts/lint_exclusions.txt
```diff
@@ -8,3 +8,4 @@
 *src/cider/type/data/sqltypes.h
 */src/cider/type/data/InlineNullValues.h
 */src/cider/common/contrib/*
+*/src/cider/exec/plan/lookup/FunctionLookupEngine.cpp
```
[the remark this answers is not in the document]
yma11: There is a legacy lint issue in this file, about an static variable definition and I am not sure about the fix.

cpp/src/cider/function/scripts/parse_ast.py
yma11: This is used for external function like `between` generation from header file. Used in CMake file.

cpp/src/cider/type/data/CiderTypes.h
yma11: Actually this content should be consolidated in other file and this file should be removed. This can be done in late refactor.

ci/scripts/lint_exclusions.txt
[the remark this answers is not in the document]
jikunshang: you can add `// NOLINT` for this issue.

cpp/src/cider/function/scalar/ExtensionFunctions.hpp
yma11: Temporarily use relative path as header can't be found if using absolute path.

harborn (approved):

PR #399, opened 2023-03-06
[POAE7-2897] string function null process optimization

jikunshang:

string function null process optimization, lower/upper optimization.

for performance

No

UT

### DIFF
--- a/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
+++ b/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
@@ -61,8 +61,6 @@ using namespace facebook::velox::functions;
 using namespace facebook::velox::exec::test;
 using namespace facebook::velox::substrait;
 
-// static const std::shared_ptr<PoolAllocator> allocator =
-//     std::make_shared<CiderDefaultAllocator>();
 namespace {
 std::pair<ArrowArray*, ArrowSchema*> veloxVectorToArrow(RowVectorPtr vec,
                                                         MemoryPool* pool) {
@@ -92,7 +90,7 @@ class StringFunctionBenchmark : public functions::test::FunctionBenchmarkBase {
     VectorFuzzer::Options opts;
     opts.vectorSize = vectorSize;
     opts.nullRatio = FLAGS_ratio;
-    opts.stringVariableLength = true;
+    opts.stringVariableLength = false;
     VectorFuzzer fuzzer(opts, pool(), FLAGS_fuzzer_seed);
 
     std::vector<VectorPtr> children;
--- a/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
+++ b/cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
@@ -45,6 +45,7 @@
 
 // This file refers velox/velox/benchmarks/basic/SimpleArithmetic.cpp
 DEFINE_int64(fuzzer_seed, 99887766, "Seed for random input dataset generator");
+DEFINE_double(ratio, 0.5, "NULL ratio in batch");
 DEFINE_int64(batch_size, 1000, "batch size for one loop");
 DEFINE_int64(loop_count, 100'000, "loop count for benchmark");
 DEFINE_bool(dump_ir, false, "dump llvm ir");
@@ -82,20 +83,31 @@ class StringFunctionBenchmark : public functions::test::FunctionBenchmarkBase {
   explicit StringFunctionBenchmark(size_t vectorSize) : FunctionBenchmarkBase() {
     functions::prestosql::registerStringFunctions();
 
-    inputType_ = ROW({{"col_str", VARCHAR()}, {"col_str2", VARCHAR()}});
+    inputType_ = ROW({{"col_str_10", VARCHAR()},
+                      {"col_str_100", VARCHAR()},
+                      {"col_str_500", VARCHAR()}});
 
     vectorSize_ = vectorSize;
     // Generate input data.
     VectorFuzzer::Options opts;
     opts.vectorSize = vectorSize;
-    // opts.nullRatio = 0.5;
-    // opts.stringVariableLength = false;
-    // opts.stringLength = 16;
+    opts.nullRatio = FLAGS_ratio;
+    opts.stringVariableLength = true;
     VectorFuzzer fuzzer(opts, pool(), FLAGS_fuzzer_seed);
 
     std::vector<VectorPtr> children;
-    children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));  // varchar
-    children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));  // varchar
+    // column 1: 10 bytes;
+    opts.stringLength = 10;
+    fuzzer.setOptions(opts);
+    children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));
+    // column 2: 100 bytes;
+    opts.stringLength = 100;
+    fuzzer.setOptions(opts);
+    children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));
+    // column 3: 500 bytes;
+    opts.stringLength = 500;
+    fuzzer.setOptions(opts);
+    children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));
 
     rowVector_ = std::make_shared<RowVector>(
         pool(), inputType_, nullptr, vectorSize, std::move(children));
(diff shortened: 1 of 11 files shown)


cpp/src/cider/exec/nextgen/context/ContextRuntimeFunctions.h
```diff
@@ -37,6 +37,14 @@ extern "C" ALWAYS_INLINE int8_t* get_query_context_string_heap_ptr(int8_t* conte
...
 }
 
+extern "C" ALWAYS_INLINE int8_t* cast_int64_to_ptr(int64_t address) {
+  return (int8_t*)address;
+}
+
+extern "C" ALWAYS_INLINE int64_t cast_ptr_to_int64(int8_t* ptr) {
+  return (int64_t)ptr;
+}
+
```
yma11, line 47: are these 2 methods similar as API `castPointerSubType()` or `castJITValuePrimitiveType`?

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -44,19 +44,17 @@ extern "C" RUNTIME_EXPORT int64_t cider_substring(const char* str, int pos, int
...
-  return pack_string_t(s);
-}
-extern "C" ALWAYS_INLINE void cider_substring_extra_ptr(char* buffer_ptr,
+extern "C" RUNTIME_EXPORT int64_t cider_substring_extra(char* string_heap_ptr,
                                                         const char* str,
                                                         int pos,
                                                         int len) {
-  memcpy(buffer_ptr, str + pos - 1, len);
+  // StringHeap* ptr = reinterpret_cast<StringHeap*>(string_heap_ptr);
+  // string_t s = ptr->addString(str + pos - 1, len);
```
yma11, line 52: will these commented lines possible be used in future? better to remove them if not.

cpp/src/cider/type/plan/StringOpExpr.cpp
```diff
@@ -316,54 +316,106 @@ JITExprValue& SubstringStringOper::codegen(CodegenContext& context) {
...
+                                        .params_vector = {ret_ptr.get()}});
+        })
+        ->build();
+    return set_expr_value(
+        arg_val.getNull(),
+        ret_len,
+        func.emitRuntimeFunctionCall(
+            "cast_int64_to_ptr",
+            JITFunctionEmitDescriptor{.ret_type = JITTypeTag::POINTER,
+                                      .params_vector = {ret_ptr_int64.get()}}));
```
yma11, line 418: Seems several `if...else...` branches are added, will it cause perf degradation if null_ratio is small and prevent future vectorization?

cpp/src/cider/exec/nextgen/context/ContextRuntimeFunctions.h
jikunshang: yes. but seems we can not do such cast via these APIs.

cpp/src/cider/type/plan/StringOpExpr.cpp
jikunshang: `if (isOutput()) {} else ` will generate different code depending on expression type, this will not cause any perf degradation. `ifBuilder` will generating two paths for null and not-null rows. Actually, it will short circuit for null path, performance is better as far as I test.  This may influence primitive type vectorization but will not for string type.

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -140,48 +136,72 @@ const uint8_t ascii_char_lower_map[] = {
     224, 225, 226, 227, 228, 229, 230, 231, 232, 233, 234, 235, 236, 237, 238, 239,
     240, 241, 242, 243, 244, 245, 246, 247, 248, 249, 250, 251, 252, 253, 254, 255};
 
+ALWAYS_INLINE void do_lower(char* out, const char* in, int str_len) {
```
bigPYJ1151, line 139: label all pointers as ```__restrict``` to eliminate overlap check code, I think there is no memory overlap in ```query_func```.

bigPYJ1151, line 139: Maybe it is unnecessary to inline such loop primitives.

cpp/src/cider/type/plan/StringOpExpr.cpp
```diff
@@ -316,54 +316,106 @@ JITExprValue& SubstringStringOper::codegen(CodegenContext& context) {
...
+              JITFunctionEmitDescriptor{
+                  .ret_type = JITTypeTag::INT32,
+                  .params_vector = {pos_param.get(),
+                                    arg_val.getLength().get(),
+                                    len_val.getValue()
+                                        ->castJITValuePrimitiveType(JITTypeTag::INT32)
+                                        .get()}});
+
+          // get string heap ptr
+          auto string_heap_ptr = func.emitRuntimeFunctionCall(
```
bigPYJ1151, line 349: I think ```JITValue``` of the string heap could be generated in ````QueryFunctionInitializer``` and saved into ```CodegenContext```.

cpp/src/cider/type/plan/StringOpExpr.cpp
```diff
@@ -387,33 +439,55 @@ JITExprValue& LowerStringOper::codegen(CodegenContext& context) {
...
+        })
+        ->build();
+
+    return set_expr_value(
+        arg_val.getNull(),
+        ret_len,
+        func.emitRuntimeFunctionCall(
+            "cast_int64_to_ptr",
+            JITFunctionEmitDescriptor{.ret_type = JITTypeTag::POINTER,
+                                      .params_vector = {ret_ptr_int64.get()}}));
```
Deegue, line 490: Q: Why we have to cast the pointer back and forth? 
It looks strange and will it lower the performance since 2 emit calls added for every method?

cpp/src/cider/exec/nextgen/context/ContextRuntimeFunctions.h
```diff
@@ -37,6 +37,14 @@ extern "C" ALWAYS_INLINE int8_t* get_query_context_string_heap_ptr(int8_t* conte
   return reinterpret_cast<int8_t*>(context_ptr->getStringHeapPtr());
 }
 
+extern "C" ALWAYS_INLINE int8_t* cast_int64_to_ptr(int64_t address) {
```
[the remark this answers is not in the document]
jikunshang: right

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -44,19 +44,15 @@ extern "C" RUNTIME_EXPORT int64_t cider_substring(const char* str, int pos, int
...
-}
-extern "C" ALWAYS_INLINE void cider_substring_extra_ptr(char* buffer_ptr,
+extern "C" RUNTIME_EXPORT int64_t cider_substring_extra(char* string_heap_ptr,
                                                         const char* str,
                                                         int pos,
                                                         int len) {
-  memcpy(buffer_ptr, str + pos - 1, len);
+  return pack_string((const int8_t*)(str + pos - 1), len);
+}
+extern "C" RUNTIME_EXPORT const char* cider_substring_extra_ptr(const char* str,
```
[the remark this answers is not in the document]
jikunshang: StringOpExpr.cpp L402

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -140,48 +136,72 @@ const uint8_t ascii_char_lower_map[] = {
     224, 225, 226, 227, 228, 229, 230, 231, 232, 233, 234, 235, 236, 237, 238, 239,
     240, 241, 242, 243, 244, 245, 246, 247, 248, 249, 250, 251, 252, 253, 254, 255};
 
+ALWAYS_INLINE void do_lower(char* out, const char* in, int str_len) {
+#pragma clang loop vectorize(enable)
```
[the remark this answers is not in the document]
jikunshang: it's compiler hints. and it's nice to have. see velox impl. https://github.com/facebookincubator/velox/blob/main/velox/functions/lib/string/StringCore.h#L91

cpp/src/cider/type/plan/StringOpExpr.cpp
jikunshang: better than what?

jikunshang: Because a pointer (of string) cannot be assigned.

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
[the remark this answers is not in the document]
Deegue: Will it be any difference or benefit if we add this line?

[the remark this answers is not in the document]
jikunshang: No.

cpp/src/cider/type/plan/StringOpExpr.cpp
Deegue: > better than what?

compare to before

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
```diff
@@ -140,48 +136,72 @@ const uint8_t ascii_char_lower_map[] = {
...
+  for (int i = 0; i < str_len; ++i) {
+    if (in[i] >= 'A' && in[i] <= 'Z') {
```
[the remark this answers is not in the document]
Deegue: `If clause` will break SIMD, I doubt this case will gain performance from SIMD.

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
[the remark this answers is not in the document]
Deegue: So, remove this line and below.

cpp/src/cider/type/plan/StringOpExpr.cpp
jikunshang: I didn't do such test. but it did better for ratio > 0 case.

cpp/src/cider-velox/benchmark/expression/StringFunctionBenchmark.cpp
```diff
@@ -82,20 +81,31 @@ class StringFunctionBenchmark : public functions::test::FunctionBenchmarkBase {
   explicit StringFunctionBenchmark(size_t vectorSize) : FunctionBenchmarkBase() {
     functions::prestosql::registerStringFunctions();
 
-    inputType_ = ROW({{"col_str", VARCHAR()}, {"col_str2", VARCHAR()}});
+    inputType_ = ROW({{"col_str_10", VARCHAR()},
+                      {"col_str_100", VARCHAR()},
+                      {"col_str_500", VARCHAR()}});
```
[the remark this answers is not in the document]
jikunshang: I think this is not necessary, our benchmark queries are binding on string length. like substring(col_str_500, 1, 500)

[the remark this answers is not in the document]
jikunshang: besides, your suggestion is google benchmark, we are using fooly benchmark here. I am not sure how to set such configurations.

cpp/src/cider/exec/nextgen/function/CiderStringFunctions.cpp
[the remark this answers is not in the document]
jikunshang: No. I will not remove this hints. I add this because I want compiler to generate vectorized instructions for this function no matter what LLVM default setting is.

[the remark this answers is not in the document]
Deegue: The point is you said there won't be any difference or benefit if we add this line, so why not remove them?

Ps. The result of your test proves the performance will be even worse if loop vectorized. (without so called ramp up)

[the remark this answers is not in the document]
jikunshang: Like I said, this is compiler hint. this may be useful someday in the future.
you mean O1 optimization? that's due to cache, or something else.  if you exchange two code blocks, the result is opposite. Please run it.

yma11 (approved):

Ma-Jian1 (approved):

PR #402, opened 2023-03-06
[POAE7-2864] Implementation of HashTable Arena and StringRef

Deegue:
1. Add `Arena.h` based on allocators, `MemoryChunk` is the memory unit. Memory is aligned to `_SC_PAGESIZE` of operation system, and it is padded to `PADDING_FOR_SIMD` for better SIMD performance and memory efficiency.
2. Add `StringRef.h` to preserve strings. It provides high performance for strings comparison through `SSE2` instruction set, which is important for HashTable when the key type is string.
3. Add `CRC32Hash` for String, which is under `SSE4.2` instruction set.
4. Add `StringRef` and `SerializedKeyHolder` to `HashTableKeyHolder` for multiple keys support.

Add Arena and StringRef.

No.

UTs.

### DIFF
--- a/cpp/src/cider/common/Arena.h
+++ b/cpp/src/cider/common/Arena.h
@@ -22,11 +22,25 @@
 
 #pragma once
 
+#include <../include/cider/CiderAllocator.h>
+#include <boost/noncopyable.hpp>
+
+#include <unistd.h>
+#include <cstdlib>
 #include <cstring>
 #include <memory>
 #include <vector>
 
-#include <boost/noncopyable.hpp>
+#if !defined(likely)
+#define likely(x) (__builtin_expect(!!(x), 1))
+#endif
+#if !defined(unlikely)
+#define unlikely(x) (__builtin_expect(!!(x), 0))
+#endif
+
+#define PADDING_FOR_SIMD 64
+
+#define NO_INLINE __attribute__((__noinline__))
 
 namespace cider::hashtable {
 
@@ -40,27 +54,126 @@ namespace cider::hashtable {
  */
 class Arena : private boost::noncopyable {
  private:
-  static size_t roundUpToPageSize(size_t s, size_t page_size) { return 0; }
+  static constexpr size_t pad_right = PADDING_FOR_SIMD - 1;
 
-  /// If MemoryChunks size is less than 'linear_growth_threshold', then use exponential
-  /// growth, otherwise - linear growth
-  ///  (to not allocate too much excessive memory).
-  size_t nextSize(size_t min_next_size) const { return 0; }
+  // Contiguous MemoryChunk of memory and pointer to free space inside it. Member of
+  // single-linked list.
+  struct alignas(16) MemoryChunk : private CiderDefaultAllocator {
+    char* begin;
+    char* pos;
+    char* end;  // does not include padding.
 
-  /// Add next contiguous MemoryChunk of memory with size not less than specified.
-  void addMemoryChunk(size_t min_size) {}
+    MemoryChunk* prev;
+
+    MemoryChunk(size_t size_, MemoryChunk* prev_) {
+      begin = reinterpret_cast<char*>(CiderDefaultAllocator::allocate(size_));
+      pos = begin;
+      end = begin + size_ - pad_right;
+      prev = prev_;
+    }
+
+    ~MemoryChunk() {
+      CiderDefaultAllocator::deallocate(reinterpret_cast<int8_t*>(begin), size());
+      delete prev;
+    }
+
+    size_t size() const { return end + pad_right - begin; }
+    size_t remaining() const { return end - pos; }
+  };
+
+  size_t growth_factor;
+  size_t linear_growth_threshold;
+
+  // Last contiguous MemoryChunk of memory.
+  MemoryChunk* head;
+  size_t size_in_bytes;
+  size_t page_size;
+
+  static size_t roundUpToPageSize(size_t s, size_t page_size) {
+    return (s + page_size - 1) / page_size * page_size;
+  }
+
+  // If MemoryChunks size is less than 'linear_growth_threshold', then use exponential
+  // growth, otherwise - linear growth
+  //  (to not allocate too much excessive memory).
+  size_t nextSize(size_t min_next_size) const {
+    size_t size_after_grow = 0;
+
+    if (head->size() < linear_growth_threshold) {
+      size_after_grow = std::max(min_next_size, head->size() * growth_factor);
+    } else {
+      // allocContinue() combined with linear growth results in quadratic
+      // behavior: we append the data by small amounts, and when it
+      // doesn't fit, we create a new MemoryChunk and copy all the previous data
+      // into it. The number of times we do this is directly proportional
+      // to the total size of data that is going to be serialized. To make
+      // the copying happen less often, round the next size up to the
+      // linear_growth_threshold.
+      size_after_grow =
+          ((min_next_size + linear_growth_threshold - 1) / linear_growth_threshold) *
+          linear_growth_threshold;
+    }
+
+    return roundUpToPageSize(size_after_grow, page_size);
+  }
+
+  // Add next contiguous MemoryChunk of memory with size not less than specified.
+  void NO_INLINE addMemoryChunk(size_t min_size) {
+    head = new MemoryChunk(nextSize(min_size + pad_right), head);
+    size_in_bytes += head->size();
+  }
+
+  friend class ArenaAllocator;
+  template <size_t>
+  friend class AlignedArenaAllocator;
 
  public:
-  /// Get piece of memory, without alignment.
-  char* alloc(size_t size) { return nullptr; }
+  explicit Arena(size_t initial_size_ = 4096,
+                 size_t growth_factor_ = 2,
+                 size_t linear_growth_threshold_ = 128 * 1024 * 1024)
+      : growth_factor(growth_factor_)
+      , linear_growth_threshold(linear_growth_threshold_)
+      , head(new MemoryChunk(initial_size_, nullptr))
+      , size_in_bytes(head->size()) {
+    page_size = sysconf(_SC_PAGESIZE);
+    if (page_size < 0) {
+      abort();
+    }
+  }
 
-  /// Get piece of memory with alignment
-  char* alignedAlloc(size_t size, size_t alignment) { return nullptr; }
+  ~Arena() { delete head; }
 
-  // TODO(Deegue): Implement and enable later
+  // Get piece of memory, without alignment.
+  char* alloc(size_t size) {
+    if (unlikely(static_cast<std::ptrdiff_t>(size) > head->end - head->pos)) {
+      addMemoryChunk(size);
+    }
+
+    char* res = head->pos;
+    head->pos += size;
+    return res;
+  }
+
+  // Get piece of memory with alignment
+  char* alignedAlloc(size_t size, size_t alignment) {
+    do {
+      void* head_pos = head->pos;
+      size_t space = head->end - head->pos;
+
+      auto* res = static_cast<char*>(std::align(alignment, size, head_pos, space));
+      if (res) {
+        head->pos = static_cast<char*>(head_pos);
+        head->pos += size;
+        return res;
+      }
+
+      addMemoryChunk(size + alignment);
+    } while (true);
+  }
+
   template <typename T>
   T* alloc() {
-    return nullptr;
+    return reinterpret_cast<T*>(alignedAlloc(sizeof(T), alignof(T)));
   }
 
   /** Rollback just performed allocation.
@@ -68,7 +181,10 @@ class Arena : private boost::noncopyable {
    * Return the resulting head pointer, so that the caller can assert that
    * the allocation it intended to roll back was indeed the last one.
    */
-  void* rollback(size_t size) { return nullptr; }
+  void* rollback(size_t size) {
+    head->pos -= size;
+    return head->pos;
+  }
 
   /** Begin or expand a contiguous range of memory.
    * 'range_start' is the start of range. If nullptr, a new range is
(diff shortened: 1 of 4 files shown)


cpp/src/cider/common/Arena.h
```diff
@@ -22,11 +22,25 @@
 
 #pragma once
 
+#include <../include/cider/CiderAllocator.h>
```
harborn, line 25: change this include path with unified style

cpp/src/cider/common/hashtable/HashTableKeyHolder.h
```diff
@@ -20,12 +20,11 @@
...
  */
 
-#pragma once
-
 #include <common/Arena.h>
+#include <common/base/StringRef.h>
 #include "type/data/funcannotations.h"
 
-namespace cider::hashtable {
+using namespace cider::hashtable;
```
harborn, line 27: not using namespace in header file.

cpp/src/cider/common/base/StringRef.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <ostream>
+
+#include "StringRef.h"
```
harborn, line 25: header file path style

cpp/src/cider/common/Arena.h
```diff
@@ -40,35 +54,137 @@ namespace cider::hashtable {
...
-  char* alloc(size_t size) { return nullptr; }
+  explicit Arena(size_t initial_size_ = 4096,
+                 size_t growth_factor_ = 2,
+                 size_t linear_growth_threshold_ = 128 * 1024 * 1024)
+      : growth_factor(growth_factor_)
+      , linear_growth_threshold(linear_growth_threshold_)
+      , head(new MemoryChunk(initial_size_, nullptr))
+      , size_in_bytes(head->size()) {
+    page_size = sysconf(_SC_PAGESIZE);
+    if (page_size < 0) {
```
harborn, line 139: throw exception?

cpp/src/cider/common/base/StringRef.h
```diff
@@ -0,0 +1,279 @@
...
+#include <nmmintrin.h>
+#include <smmintrin.h>
+
+namespace cider::hashtable {
+
+/**
+ * The std::string_view-like container to avoid creating strings to find substrings in the
+ * hash table.
+ */
+struct StringRef {
```
harborn, line 45: I think ``` class StringRef``` will be better

cpp/src/cider/common/Arena.h
```diff
@@ -40,35 +54,137 @@ namespace cider::hashtable {
  */
 class Arena : private boost::noncopyable {
  private:
-  static size_t roundUpToPageSize(size_t s, size_t page_size) { return 0; }
+  static constexpr size_t pad_right = PADDING_FOR_SIMD - 1;
+
+  // Contiguous MemoryChunk of memory and pointer to free space inside it. Member of
+  // single-linked list.
+  struct alignas(16) MemoryChunk : private CiderDefaultAllocator {
+    char* begin;
```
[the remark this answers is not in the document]
jikunshang: class member should be named with underscore. https://google.github.io/styleguide/cppguide.html#Variable_Names

cpp/src/cider/common/Arena.h
```diff
@@ -40,35 +54,137 @@ namespace cider::hashtable {
...
+      end = begin + size_ - pad_right;
+      prev = prev_;
+    }
+
+    ~MemoryChunk() {
+      CiderDefaultAllocator::deallocate(reinterpret_cast<int8_t*>(begin), size());
+      delete prev;
+    }
+
+    size_t size() const { return end + pad_right - begin; }
```
jikunshang, line 80: why not use a class member to track size?

cpp/src/cider/common/base/StringRef.h
```diff
@@ -0,0 +1,279 @@
...
+#include <common/contrib/cityhash102/include/city.h>
+
+#include <emmintrin.h>
+#include <nmmintrin.h>
+#include <smmintrin.h>
+
+namespace cider::hashtable {
+
+/**
+ * The std::string_view-like container to avoid creating strings to find substrings in the
```
jikunshang, line 42: why need to find substring in hashtable?

cpp/src/cider/common/Arena.h
```diff
@@ -40,35 +54,137 @@ namespace cider::hashtable {
  */
 class Arena : private boost::noncopyable {
```
jikunshang, line 55: what's the relationship between `Arena`, `MemoryChunk`, `CiderXXAllocator`. From your description, `Arena` is an Allocator, while it didn't extend `CiderAllocator`, nor accept a customized Allocator. And if it's a allocator, please rename it as `ArenaAllocator`.
Besides, please create a cpp file to hide implementation.

cpp/src/cider/common/Arena.h
```diff
@@ -85,38 +201,87 @@ class Arena : private boost::noncopyable {
...
 
   /// Bad method, don't use it -- the MemoryChunks are not your business, the entire
   /// purpose of the arena code is to manage them for you, so if you find
   /// yourself having to use this method, probably you're doing something wrong.
-  size_t remainingSpaceInCurrentMemoryChunk() const { return 0; }
+  size_t remainingSpaceInCurrentMemoryChunk() const { return head->remaining(); }
 };
 
+using ArenaPtr = std::shared_ptr<Arena>;
+using Arenas = std::vector<ArenaPtr>;
```
jikunshang, line 285: why will use this?

cpp/src/cider/common/hashtable/HashTableKeyHolder.h
```diff
@@ -20,12 +20,11 @@
  * under the License.
  */
 
-#pragma once
```
jikunshang, line 23: why remove this.

cpp/src/cider/common/base/StringRef.h
```diff
@@ -0,0 +1,279 @@
...
+    case 2:
+      if (!compareSSE2(p1 + 16, p2 + 16))
+        return false;
+      [[fallthrough]];
+    case 1:
+      if (!compareSSE2(p1, p2))
+        return false;
+  }
+
+  return compareSSE2(p1 + size - 16, p2 + size - 16);
```
jikunshang, line 158: `compareSSE2` will always compare 16 bytes? what will happen if call `memequalSSE2Wide(p1, p2, 20)`?

Deegue: Yes, `compareSSE2` will be called twice for a string of 20 bytes.

cpp/src/cider/common/Arena.h
Deegue: make it easy to use

cpp/src/cider/common/base/StringRef.h
jikunshang: oh, it will compare [0,15] and [4,19], right?

cpp/src/cider/common/base/StringRef.h
Deegue: If it's not under `cider::hashtable` namespace, shall we use this class to handle all string ops like substring?

cpp/src/cider/common/base/StringRef.h
Deegue: yes

cpp/src/cider/common/base/StringRef.h
harborn: no difference with std::string_view?

cpp/src/cider/common/base/StringRef.h
```diff
@@ -0,0 +1,279 @@
...
+  std::string toString() const { return std::string(data, size); }
+  explicit operator std::string() const { return toString(); }
+
+  std::string_view toView() const { return std::string_view(data, size); }
+  constexpr explicit operator std::string_view() const {
+    return std::string_view(data, size);
+  }
+};
+
+using StringRefs = std::vector<StringRef>;
```
harborn, line 76: this alias not used

cpp/src/cider/common/Arena.h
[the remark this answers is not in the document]
Deegue: _"struct members not have the trailing underscores that data members in classes have."_

cpp/src/cider/common/Arena.h
Deegue: reduce extra memory cost

cpp/src/cider/common/base/StringRef.h
```diff
@@ -0,0 +1,279 @@
...
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(p1) + 1),
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(p2) + 1))),
+             _mm_and_si128(
+                 _mm_cmpeq_epi8(
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(p1) + 2),
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(p2) + 2)),
+                 _mm_cmpeq_epi8(
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(p1) + 3),
+                     _mm_loadu_si128(reinterpret_cast<const __m128i*>(p2) + 3)))));
+}
```
harborn, line 104: does this function has implemented in xsimd?

cpp/src/cider/common/hashtable/HashTableKeyHolder.h
```diff
@@ -89,4 +90,52 @@ inline void ALWAYS_INLINE keyHolderPersistKey(Key&&) {}
...
+
+namespace cider::hashtable {
+
+/** SerializedKeyHolder is a key holder for a StringRef key that is already
+ * serialized to an Arena. The key must be the last allocation in this Arena,
+ * and is discarded by rolling back the allocation.
+ */
+struct SerializedKeyHolder {
+  StringRef key;
+  Arena& pool;
```
harborn, line 125: using pointer?

cpp/src/cider/common/Arena.h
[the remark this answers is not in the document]
jikunshang: https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes 
Use a struct only for passive objects that carry data; everything else is a class.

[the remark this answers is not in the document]
Deegue: Just like the guide tells, I think it's ok for here to use struct since `MemoryChunk` doesn't need much _"functionality or invariants"_.

cpp/src/cider/common/base/StringRef.h
Deegue: It differs when the size of string changes. In substring cases, what we need is to change the size and keep the data pointer instead of creating a new string.

cpp/src/cider/common/base/StringRef.h
```diff
@@ -0,0 +1,279 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <cassert>
+#include <functional>
+#include <iosfwd>
+#include <stdexcept>
```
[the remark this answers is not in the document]
Deegue: removed

cpp/src/cider/common/hashtable/HashTableKeyHolder.h
Deegue: done

cpp/src/cider/common/Arena.h
Deegue: done

cpp/src/cider/common/Arena.h
```diff
@@ -40,35 +54,137 @@ namespace cider::hashtable {
  */
 class Arena : private boost::noncopyable {
  private:
-  static size_t roundUpToPageSize(size_t s, size_t page_size) { return 0; }
+  static constexpr size_t pad_right = PADDING_FOR_SIMD - 1;
+
+  // Contiguous MemoryChunk of memory and pointer to free space inside it. Member of
+  // single-linked list.
+  struct alignas(16) MemoryChunk : private CiderDefaultAllocator {
```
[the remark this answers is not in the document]
Deegue: Put them together because `MemoryChunk` is only used in `Arena`.

cpp/src/cider/common/hashtable/HashTableKeyHolder.h
Deegue: The lifecycle of pool here should be along with `SerializedKeyHolder`

cpp/src/cider/common/Arena.h
Deegue: description editted

cpp/src/cider/common/base/StringRef.h
Deegue: am not clear about xsimd, but I think all `SSE` instructions can be replaced by avx2 except a small part which is not supported or deprecated.

cpp/src/cider/common/hashtable/HashTableKeyHolder.h
Deegue: I think we should use the value of `SerializedKeyHolder` instead of the pointer, and `SerializedKeyHolder` is a struct type.

cpp/src/cider/common/Arena.h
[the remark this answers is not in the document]
jikunshang: Any suggestion on this @Ma-Jian1

jikunshang (approved):

cpp/src/cider/common/base/StringRef.cpp
harborn, line 25: #include "common/base/StringRef.h"

cpp/src/cider/common/Arena.h
harborn, line 55: Will Arena be used under multi thread case?

cpp/src/cider/common/Arena.h
```diff
@@ -40,35 +54,137 @@ namespace cider::hashtable {
...
+    ~MemoryChunk() {
+      CiderDefaultAllocator::deallocate(reinterpret_cast<int8_t*>(begin), size());
+      delete prev;
+    }
+
+    size_t size() const { return end + pad_right - begin; }
+    size_t remaining() const { return end - pos; }
+  };
+
+  size_t growth_factor;
```
harborn, line 84: default 2 is not very good, I think 1.5 will be better.

cpp/src/cider/common/Arena.h
```diff
@@ -40,35 +54,137 @@ namespace cider::hashtable {
...
 
  public:
-  /// Get piece of memory, without alignment.
-  char* alloc(size_t size) { return nullptr; }
+  explicit Arena(size_t initial_size_ = 4096,
+                 size_t growth_factor_ = 2,
+                 size_t linear_growth_threshold_ = 128 * 1024 * 1024)
+      : growth_factor(growth_factor_)
+      , linear_growth_threshold(linear_growth_threshold_)
+      , head(new MemoryChunk(initial_size_, nullptr))
```
harborn, line 136: when will release head's resource?

Deegue: Memory will be reused if we don't need the objects anymore, so it won't be released until the task is completed.

cpp/src/cider/common/Arena.h
Deegue: It should be only single thread in our cases.

cpp/src/cider/common/base/StringRef.cpp
Deegue: It will be simpler if two files are under the same folder?

cpp/src/cider/common/Arena.h
Deegue: I think 2 is a good growth factor since we set the initial size of memory is a multiple of 2. https://github.com/intel/BDTK/blob/3f7fde97ba58a3bdf904a176f1671618a92010e5/cpp/src/cider/common/hashtable/HashTable.h#L254

harborn (approved):

PR #407, opened 2023-03-08
[POAE7-2909] Add std::map, std::unordered_map and folly::F14FastMap to HashTable benchmark

Deegue:
Add std::map, std::unordered_map and folly::F14FastMap to the benchmark of HashTable.
The performance will be better if we use more efficient allocator in the future.

The conclusion is cider optimized HashTable for int8 and int16 gains
**10x-16x** performance better than folly::F14FastMap,
**15x-25x** better than cider basic HashTable,
**36x-100x** better than std::map.

The result shows below:

image

Improvement.

No.

UTs.

### DIFF
--- a/cpp/src/cider/benchmarks/HashMapBenchmark.cpp
+++ b/cpp/src/cider/benchmarks/HashMapBenchmark.cpp
@@ -27,6 +27,8 @@
 #include <common/hashtable/HashMap.h>
 #include <common/hashtable/HashTable.h>
 #include <common/hashtable/HashTableAllocator.h>
+#include <folly/container/F14Map.h>
+
 #include <tests/utils/Utils.h>
 
 using namespace cider::hashtable;
@@ -42,19 +44,19 @@ template <typename KeyType, typename Map>
 int32_t NO_INLINE bench(const std::vector<KeyType>& data) {
   Map map;
 
-  for (auto value : data) {
+  for (auto key : data) {
     typename Map::LookupResult it;
     bool inserted;
 
-    map.emplace(value, it, inserted);
+    map.emplace(key, it, inserted);
     if (inserted)
       it->getMapped() = 1;
     else
       ++it->getMapped();
   }
 
-  for (auto value : data) {
-    auto it = map.find(value);
+  for (auto key : data) {
+    auto it = map.find(key);
     auto curr = ++it;
     if (curr)
       curr->getMapped();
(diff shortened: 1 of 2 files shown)


cpp/src/cider/benchmarks/HashMapBenchmark.cpp
```diff
@@ -114,6 +116,59 @@ static std::tuple<std::vector<T>, std::vector<bool>> generateAndFillVector(
...
+  std::vector<KeyType> data = prepare_data<KeyType>(row_num);
+  for (auto _ : state) {
+    int32_t collisions = bench<KeyType, HashTableType>(data);
+    state.counters["Collisions"] = collisions;
+  }
+}
+
+template <typename KeyType>
+void NO_INLINE bench_std(const std::vector<KeyType>& data) {
+  std::map<KeyType, int8_t> map;
```
jikunshang, line 128: please use `std::unordered_map` instead.

jikunshang (commented):
comapre to https://github.com/intel/BDTK/pull/377, why collision of baseline hash table become 0?

cpp/src/cider/benchmarks/HashMapBenchmark.cpp
Deegue: Keep `std::map` in this pr and add `std::unordered_map` later.

Deegue:
> comapre to #377, why collision of baseline hash table become 0?

Initial size of basic map is changed to 1<<12.

cpp/src/cider/benchmarks/HashMapBenchmark.cpp
jikunshang: `std::map` is not a hash map implementation. feel free to keep it here if you think it's necessary.

Deegue: It's necessary since `std::map` is something like `TreeMap` in java, which doesn't use hash function but will be more efficient than `std::unordered_map` in many cases.

Deegue:
Besides, changing hash function to `crc32hash` will also improve the performance of basic hashtable and reduce collisions and CPU time.

jikunshang (approved):

cpp/src/cider/benchmarks/HashMapBenchmark.cpp
harborn: It seems this function does nothing for input data
for benchmark, you should use some info of map to avoid this function be optimized nothing executed under O3
such like ``` return map.size();```

harborn: maybe google benchmark will stop compiler optimize this function.

PR #412, opened 2023-03-14
[POAE7-2914]OutputColumnVar supports auto-vectorization

Ma-Jian1:
1. OutputColumnVar supports auto-vectorization
2. memory optimization for arrow array and schema

perf

No

benchmark

INFRA

### DIFF
--- a/cpp/src/cider/type/plan/ColumnExpr.h
+++ b/cpp/src/cider/type/plan/ColumnExpr.h
@@ -127,13 +127,18 @@ class ColumnVar : public Expr {
 class OutputColumnVar : public Expr {
  public:
   explicit OutputColumnVar(const std::shared_ptr<ColumnVar>& col)
-      : Expr(col->get_type_info()), col_(col) {}
+      : Expr(col->get_type_info()), col_(col) {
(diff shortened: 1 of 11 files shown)


cpp/src/cider/type/plan/ColumnExpr.h
```diff
@@ -127,13 +127,18 @@ class ColumnVar : public Expr {
...
+    initAutoVectorizeFlag();
+  }
 
   JITExprValue& codegen(CodegenContext& context) override;
 
   ExprPtrRefVector get_children_reference() override { return {&col_}; }
 
  private:
+  void initAutoVectorizeFlag() {
+    auto_vectorizable_ = isVectorizableType(type_info.get_type());
```
yma11, line 140: When this property is used?

bigPYJ1151 (approved):

cpp/src/cider/type/plan/ColumnExpr.h
Ma-Jian1: transformer split expressions into groups

yma11 (approved):

PR #427, opened 2023-03-28
[POAE7-2930]  fix query_function core dump issue when integrate with JNI

chaojun-zhang:

 load cider_function.so when init LLVMJitEngine

fix query_function core dump issue when integrate with JNI

No

Passed existing UT

cider , jni

### DIFF
--- a/cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITEngine.cpp
+++ b/cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITEngine.cpp
@@ -21,11 +21,27 @@
 
 #include <llvm/Analysis/TargetTransformInfo.h>
 #include <llvm/ExecutionEngine/JITEventListener.h>
+#include <llvm/Support/DynamicLibrary.h>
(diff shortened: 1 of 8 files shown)


cpp/src/cider/exec/nextgen/jitlib/llvmjit/LLVMJITEngine.cpp
```diff
@@ -21,11 +21,27 @@
...
+#include <filesystem>
 
 #include "exec/nextgen/jitlib/llvmjit/LLVMJITEngine.h"
 #include "exec/nextgen/jitlib/llvmjit/LLVMJITModule.h"
+#include "util/filesystem/cider_path.h"
 
 namespace cider::jitlib {
+
+namespace {
+static void loadCiderFunction() {
```
jikunshang, line 34: `loadCiderFunctionLibrary` or `loadLibrary(const std::string& lib_name)`

chaojun-zhang: Updated

PR #431, opened 2023-03-30
[POAE7-2931] Add benchmark for StringHashTable

Deegue:
Add benchmark of HashTable for key type string.
This benchmark generates 10k, 100k, 1M rows of data with low cardinality (~62), and run on `std::map`, `std::unordered_map`, `folly:F14FastMap` and `cider::optimized_map`.

Short string: (length = 10)

BM_s_str1

Long string: (length =500)

BM_l_str1

Improvement.

No.

UTs

### DIFF
--- a/cpp/src/cider/benchmarks/StringHashMapBenchmark.cpp
+++ b/cpp/src/cider/benchmarks/StringHashMapBenchmark.cpp
@@ -1,0 +1,184 @@
+/*
+ * Copyright(c) 2022-2023 Intel Corporation.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


cpp/src/cider/benchmarks/StringHashMapBenchmark.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+size_t s_str_len = 10;
+size_t l_str_len = 500;
+
+#define N_MAX std::numeric_limits<T>::max()
+#define N_MIN std::numeric_limits<T>::min()
+
+#define NO_INLINE __attribute__((__noinline__))
+
+static std::string sequence_string(size_t length, size_t index) {
```
jikunshang, line 48: this function will generate fixed string rather than random string? if so hash map will only store (10+26+26) unique value?

Deegue: yes

harborn (approved):

PR #434, opened 2023-04-04
[POAE7-2935] copy elimination for some bare columns

Ma-Jian1:
copy elimination for some bare columns

In `select a, b, b+c from test`, column `a` is not toughed in the project, column `b` is not modified,
these columns should not allocate new memory and copy from input to output.
This pr we skip columns `a` and `b` in the project and just copy the children pointer in `resetBatch` and clear
input children pointer in the `LazyNode` to ensure these columns are not released.

Something needs to be improved:
1. maybe we can remove `LazyNode`, and just do the work in the deferred function, then the StageNode can be removed.
2. we just support copy elimination in the project, skipped if has others op(eg. filter, join)
3. `resetBatch` should match input reuse(ie. if we reuse the input, resetBatch must be passed the input array and schema)
4. project op should be removed in `select a, b from test`, just keep the lazy op

performance optimization

No

UT and benchmark

INFRA


cpp/src/cider-velox/benchmark/expression/PipelineOperator.cpp
```diff
@@ -293,11 +296,16 @@ BENCHMARK_RELATIVE(nextgenOpt) {
...
+  // warmup: preload YAML
+  FunctionLookupEngine::getInstance(PlatformType::PrestoPlatform);
+
   benchmark = std::make_unique<PipelineOperator>();
   // mimic tpch q1 project expression
   benchmark->initialize(
       {"c", "c_1", "d", "d_1", "d_2", "d_3"},
       {"c", "c_1", "d", "d_1", "d_1*(1.0-d_2)", "d_1*(1.0-d_2)*(1.0+d_3)", "d_2"});
+  // {"i64", "i64_1", "d", "d_1", "d_2", "d_3"},
+  // {"i64", "i64_1", "d", "d_1", "d_1*(1.0-d_2)", "d_1*(1.0-d_2)*(1.0+d_3)", "d_2"});
```
bigPYJ1151, line 308: Such code could be removed if unused.

cpp/src/cider-velox/benchmark/tpch/TpchInMemBuilder.cpp
```diff
@@ -151,7 +151,8 @@ TpchPlan TpchInMemBuilder::getQ1Plan(double scaleFactor) const {
           //   .tableScan(kLineitem, selectedRowType, fileColumnNames, {filter})
           .tableScan(Table::TBL_LINEITEM, std::move(selectedColumns), scaleFactor)
           .capturePlanNodeId(lineitemPlanNodeId)
-          .filter(filter)
+          // velox filter will wrap result childrens in dictionary
+          //   .filter(filter)
```
bigPYJ1151, line 155: Comment filters for debugging? Better to revert it if so.

cpp/src/cider/exec/nextgen/operators/OpNode.h
```diff
@@ -57,7 +56,7 @@ class OpNode : public std::enable_shared_from_this<OpNode> {
   /// \brief The name of the operator node
   const char* name() const { return name_; }
 
-  void setInputOpNode(const OpNodePtr& input) { input_ = input; }
+  // void setInputOpNode(const OpNodePtr& input) { input_ = input; }
```
jikunshang, line 59: remove?

jikunshang (approved):

cpp/src/cider-velox/benchmark/tpch/TpchInMemBuilder.cpp
Ma-Jian1: copy elimination doesn't support filter. So I commend all filter operations and just benchmark project op.

jikunshang (approved):

bigPYJ1151 (approved):

